PR #1, opened 2026-01-06
Network switch

CCP-2FaceMonkey:
## Summary

Network switching feature with per-network authentication and device data isolation.

### Network Switching
- Added network selector UI for switching between Sui devnet/testnet
- Per-network JWT storage (separate login sessions per network)
- Per-network device data (nonce, maxEpoch, jwtRandomness stored per-chain)
- Seamless switching when already authenticated on target network
- Automatic rollback to previous network on login failure

### Bug Fixes
- Fixed `maxEpoch: null` error when switching networks (reactive subscription in `useDevice`)
- Fixed "JWT nonce doesn't match device nonce" error (prevented premature device data regeneration)
- Fixed stale device data after login by rehydrating deviceStore from Chrome storage
- Fixed wallet lock on logout

### Testing
- Added `useDevice.test.tsx` - reactive chain subscription tests
- Added `networkStore.test.ts` - network switching and JWT validation tests
- Added lock/logout tests for keeper and auth stores

FlowChart:

## Network Switching Flow
```mermaid
flowchart TD
    A[User clicks Switch to Testnet] --> B{checkNetworkSwitch}
    B -->|Has JWT| C[Seamless Switch]
    B -->|No JWT| D[Show Sign In Required Dialog]
    C --> E[setChain testnet]
    E --> F[Done - Switch Complete]
    D --> G[User clicks Sign In]
    G --> H[forceSetChain testnet]
    H --> I[login function called]
    I --> J[Background Handler]
    J --> K[initializeForChain testnet]
    K --> L[Updates Chrome Storage<br/>with new nonce, maxEpoch]
    L --> M[OAuth Flow Completes]
    M --> N[JWT Stored for testnet]
    N --> O[rehydrateDeviceStore]
    O --> P[Popup reads updated<br/>networkData from Chrome]
    P --> Q[useDevice returns<br/>correct maxEpoch]
    Q --> R[Test Transaction Works]
  ```

### DIFF
--- a/packages/shared/src/stores/deviceStore.ts
+++ b/packages/shared/src/stores/deviceStore.ts
@@ -469,9 +469,10 @@ export const useDeviceStore = create<DeviceState>()(
         // Epoch start is a Unix timestamp in milliseconds
         const { epoch, epochDurationMs, epochStartTimestampMs } =
           await suiClient.getLatestSuiSystemState();
-        const numericMaxEpoch = Number(epoch); // Set to current epoch for now, can increase validity window in the future
+        // Use current epoch - must stay aligned with maxEpochTimestampMs below
+        const numericMaxEpoch = Number(epoch);
 
-        // 4. Set maxEpoch expiry (aligned with numericMaxEpoch which is current epoch)
+        // 4. Set maxEpoch expiry - must stay aligned with numericMaxEpoch (current epoch)
         const maxEpochTimestampMs =
           Number(epochStartTimestampMs) + Number(epochDurationMs);
 
@@ -632,6 +633,21 @@ export const useDeviceStore = create<DeviceState>()(
       },
 
       unlock: async (pin: string) => {
+        // Helper to set unlocked state with optional public key
+        const setUnlockedState = (publicKey: PublicKey | null) => {
+          if (publicKey) {
+            set({
+              isLocked: false,
+              error: null,
+              ephemeralPublicKey: publicKey,
+              ephemeralPublicKeyBytes: Array.from(publicKey.toRawBytes()),
+              ephemeralPublicKeyFlag: publicKey.flag(),
+            });
+          } else {
+            set({ isLocked: false, error: null });
+          }
+        };
+
         try {
           const storedKey = get().ephemeralKeyPairSecretKey;
 
@@ -651,18 +667,7 @@ export const useDeviceStore = create<DeviceState>()(
 
             // Unlock the vault with PIN (decrypts the stored key)
             const publicKey = await ephKeyService.unlockVault(null, pin);
-
-            if (publicKey) {
-              set({
-                isLocked: false,
-                error: null,
-                ephemeralPublicKey: publicKey,
-                ephemeralPublicKeyBytes: Array.from(publicKey.toRawBytes()),
-                ephemeralPublicKeyFlag: publicKey.flag(),
-              });
-            } else {
-              set({ isLocked: false, error: null });
-            }
+            setUnlockedState(publicKey);
             return;
           }
 
@@ -673,18 +678,7 @@ export const useDeviceStore = create<DeviceState>()(
           }
 
           const publicKey = await ephKeyService.unlockVault(storedKey, pin);
-
-          if (publicKey) {
-            set({
-              isLocked: false,
-              error: null,
-              ephemeralPublicKey: publicKey,
-              ephemeralPublicKeyBytes: Array.from(publicKey.toRawBytes()),
-              ephemeralPublicKeyFlag: publicKey.flag(),
-            });
-          } else {
-            set({ isLocked: false, error: null });
-          }
+          setUnlockedState(publicKey);
         } catch (error) {
           log.error("Error decrypting secret key", error);
           set({
--- a/apps/extension/src/lib/background/handlers/authHandlers.ts
+++ b/apps/extension/src/lib/background/handlers/authHandlers.ts
@@ -140,7 +140,7 @@ async function handleExtLogin(
   // Check if the keeper has an unlocked ephemeral key
   let keeperStatus = await checkKeeperUnlocked();
   if (!keeperStatus.unlocked) {
-    // If device data exists, retry once (unlock might be in progress)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: apps/extension/src/features/wallet/hooks/useTestTransaction.ts [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+      if (!ephemeralPublicKey) {
+        throw new Error("Ephemeral public key not found");
+      }
+
+      const tx = new Transaction();
+      tx.setSender(user.profile?.sui_address as string);
+      const txb = await tx.build({ client: suiClient });
+
+      const { bytes, zkSignature } = await zkSignAny("TransactionData", txb, {
+        user,
```
ccpsondheim: With network switching, some values of `user` now need to be derived from the stored jwt instead of the oidc user. The stored oidc `user` will always be the last user that was signed in; ie if [1] you signed into devnet, then [2] signed into testnet, then [3] switched back into devnet, then the stored oidc user would still remain the one on testnet, alongside having the testnet nonce. This causes submitted devnet transactions to then fail.

packages/shared/src/stores/deviceStore.ts
```diff
@@ -130,6 +133,14 @@ export const useDeviceStore = create<DeviceState>()(
         return get().networkData[currentChain]?.nonce ?? null;
       },
 
+      getJwtRandomness: (chain?: SuiChain) => {
+        const currentChain = chain || useNetworkStore.getState().chain;
+        // Prefer per-network jwtRandomness, fallback to global for backwards compatibility
```
ccpsondheim: It's not a good idea to have the global fallback as we know that fallbacks will definitely fail if we attempt passing transactions with it. A correct `jwtrandomness` is a dependency of a transaction passing, so any error handling should happen prior to reaching this point.

packages/shared/src/stores/deviceStore.ts
```diff
@@ -174,7 +195,41 @@ export const useDeviceStore = create<DeviceState>()(
...
+                // BUT: Only regenerate if we don't have a JWT with matching nonce
+                // (regenerating would cause nonce mismatch)
                 if (!nonce || !maxEpoch || !maxEpochTimestampMs) {
+                  // Check if we have a JWT for this network
+                  const hasJwt = await hasJwtForNetwork(currentChain);
+                  if (hasJwt) {
+                    // We have a JWT - check if nonce matches
+                    const { getJwtForNetwork } = await import(
+                      "../auth/storageService"
+                    );
```
ccpsondheim: Can this import be at the top of the file to align with the rest of the imports?

packages/shared/src/stores/deviceStore.ts
```diff
@@ -259,8 +318,14 @@ export const useDeviceStore = create<DeviceState>()(
                 }
 
                 if (persistedDeviceStoreState) {
-                  jwtRandomness =
-                    persistedDeviceStoreState.jwtRandomness ?? jwtRandomness;
+                  // Get jwtRandomness from persisted network data or global (for backwards compatibility)
```
ccpsondheim: See comment above re: global jwtRandomness

packages/shared/src/stores/deviceStore.ts
```diff
@@ -382,21 +485,27 @@ export const useDeviceStore = create<DeviceState>()(
         const maxEpochTimestampMs =
           Number(epochStartTimestampMs) + Number(epochDurationMs) * 2;
```
ccpsondheim: This should match `numericMaxEpoch`, which is set to the current epoch for now (ie not `+ Number(epochDurationMs) * 2`)

packages/shared/src/types/stores.ts
```diff
@@ -21,6 +21,7 @@ export interface NetworkDataEntry {
   nonce: string | null;
   maxEpoch: string | null;
   maxEpochTimestampMs: number | null;
+  jwtRandomness: string | null; // Per-network jwtRandomness to prevent cross-network nonce conflicts
```
ccpsondheim: Might be excessive commenting

ccpsondheim (commented):
Overall great, just a few comments

packages/shared/src/stores/deviceStore.ts
CCP-2FaceMonkey: Added comment explaining dynamic import is required to avoid circular dependency: deviceStore → auth → authStore → deviceStore.

apps/extension/src/features/wallet/components/SignTransaction.tsx
```diff
@@ -11,7 +11,7 @@ import { createSuiClient } from "@evevault/shared/sui";
 import type { PendingTransaction } from "@evevault/shared/types";
 import { buildTx, createLogger } from "@evevault/shared/utils";
 import { zkSignAny } from "@evevault/shared/wallet";
-import type { TransactionData } from "@mysten/sui/transactions";
+import { Transaction, type TransactionData } from "@mysten/sui/transactions";
```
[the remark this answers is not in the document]
ccpsondheim: Ignore - this import is a bugfix

packages/shared/src/stores/deviceStore.ts
ccpsondheim: This is great, but i'm noticing that there are some places where `deviceStore.jwtRandomness` is still being called such as in `authHandlers.ts L212`. These should be updated.

packages/shared/src/stores/deviceStore.ts
```diff
@@ -378,25 +471,27 @@ export const useDeviceStore = create<DeviceState>()(
           await suiClient.getLatestSuiSystemState();
         const numericMaxEpoch = Number(epoch); // Set to current epoch for now, can increase validity window in the future
 
-        // 4. Set maxEpoch expiry
+        // 4. Set maxEpoch expiry (aligned with numericMaxEpoch which is current epoch)
```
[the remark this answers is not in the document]
ccpsondheim: This shouldn't be relaxed in the future - the timestamp should always be aligned with numericMaxEpoch. The last comment bullet point should be ignored

PR #6, opened 2026-01-13
feat: test patch user nonce, jwt refresh

fcsondheim:
retrieve updated nonce when refreshing jwt
patch nonce from browser to fusionauth 3s before refresh req is sent


CCP-2FaceMonkey (commented):
Nitpicks and thoughts about this new temp service.

CCP-2FaceMonkey (approved):
LGTM

PR #26, opened 2026-02-19
feat: start integrating quasar http proxy

fcsondheim:
- upgrade `@mysten/sui` to v2
- start integrating sponsored tx method to quasar
- add remaining features to wallet on connect

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,8 @@
+# EVE Vault – Claude Code Guidelines
+
+## Git Workflow
+- **All changes must go through pull requests** — direct pushes to `main` are blocked by branch protection
+- **All PRs should be squash merged**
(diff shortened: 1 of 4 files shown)


apps/extension/src/lib/background/handlers/walletHandlers.ts
```diff
@@ -95,29 +103,81 @@ async function handleApprovePopup(
...
-      action: message.action,
-      assembly: message.assembly,
-      chain: message.chain,
+      action,
+      assembly,
+      chain,
     });
 
+    // Fetch the txb to be signed from the Quasar proxy
+    const response = await fetch(
```
CCP-2FaceMonkey: why use fetch? not react-query?

ccpsondheim: React query would fail here as this is outside of a React component and just in the background script.

PR #42, opened 2026-03-08
Feat: Implements device reset and centralized logout

CCP-2FaceMonkey:
This release introduces the ability for users to reset their EVE Vault data on a specific device, addressing instances where a user might forget their PIN. It also refactors the logout process to be more robust and consistent across web and extension environments.

- **Enables Device Reset:** Adds a "Forgot PIN" option to the Lock Screen, which, after confirmation, performs a full device reset. This involves clearing all local storage, ephemeral key pairs, JWTs, and resetting in-memory stores, ensuring a fresh start for the user without affecting their funds.
- **Centralizes FusionAuth Logout:** Implements a dedicated utility (`redirectToFusionAuthLogout`) to handle IdP session termination. This function ensures the user is properly logged out from FusionAuth, improving security and consistency for both web applications and browser extensions.
- **Enhances OAuth Callback Stability:** Introduces a guard in the OAuth callback flow to prevent redundant code exchanges. This ensures that the authentication process is stable, particularly in strict development modes or during page reloads.
- **Improves User Experience:** Updates Lock Screen components to integrate seamlessly with the new reset flow, and refines modal styling for better user interface consistency.
- **Refines Transaction Handling:** Enhances the handling of transaction results in the Wallet Screen to explicitly check for failed transactions, leading to more accurate error reporting and digest retrieval.

### DIFF
--- a/packages/shared/src/auth/resetVaultOnDevice.ts
+++ b/packages/shared/src/auth/resetVaultOnDevice.ts
@@ -1,16 +1,9 @@
-import {
-  SUI_DEVNET_CHAIN,
-  SUI_LOCALNET_CHAIN,
-  SUI_MAINNET_CHAIN,
-  SUI_TESTNET_CHAIN,
-  type SuiChain,
-} from "@mysten/wallet-standard";
+import { SUI_TESTNET_CHAIN } from "@mysten/wallet-standard";
 import { ephKeyService, zkProofService } from "../services/vaultService";
 import { useDeviceStore } from "../stores/deviceStore";
 import { useNetworkStore } from "../stores/networkStore";
 import { useTokenListStore } from "../stores/tokenListStore";
 import { DEFAULT_TOKENS_BY_CHAIN } from "../types/networks";
-import type { NetworkDataEntry } from "../types/stores";
 import {
   cleanupExtensionStorage,
   cleanupOidcStorage,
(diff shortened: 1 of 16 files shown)


packages/shared/src/auth/resetVaultOnDevice.ts
```diff
@@ -0,0 +1,138 @@
...
+const log = createLogger();
+
+function createEmptyNetworkDataEntry(): NetworkDataEntry {
+  return {
+    nonce: null,
+    maxEpoch: null,
+    maxEpochTimestampMs: null,
+    jwtRandomness: null,
+  };
+}
```
ccpsondheim: This is duplicated in deviceStore.ts ln 82. Can it be imported?

packages/shared/src/auth/resetVaultOnDevice.ts
```diff
@@ -0,0 +1,138 @@
...
+    loading: false,
+    error: null,
+  });
+
+  const emptyNetworkData: Partial<Record<SuiChain, NetworkDataEntry>> = {
+    [SUI_DEVNET_CHAIN]: createEmptyNetworkDataEntry(),
+    [SUI_TESTNET_CHAIN]: createEmptyNetworkDataEntry(),
+    [SUI_LOCALNET_CHAIN]: createEmptyNetworkDataEntry(),
+    [SUI_MAINNET_CHAIN]: createEmptyNetworkDataEntry(),
+  };
```
ccpsondheim: This is duplicated in device store; the store is initialised with empty values. Can we centralise this logic so that it's reused across these files?

CCP-2FaceMonkey:
> some places where code can be reused

yes thanks, updated

ccpsondheim (approved):
lgtm

PR #58, opened 2026-03-11
Fix/graphqlerror

CCP-2FaceMonkey:


ccp-raudur:
Two review concerns from the diff:

1. This PR currently mixes a narrow entrypoint fix with workspace-wide runtime upgrades (`react`, `react-dom`, and `@tanstack/react-query` in the root, web, extension, and shared `package.json`s). Even if those version bumps are valid, bundling them into `fix/graphqlerror` makes regression attribution much harder and widens the review surface well beyond the four entrypoints that changed behavior. If the upgrades are the actual root cause fix, it would help to call that out explicitly; otherwise I’d strongly prefer splitting them into a separate PR.

2. The new `QueryClientProvider` and `ToastProvider` wrappers in the signing entrypoints don’t appear to have a clear local consumer in the mounted signing flows. From the surrounding code, `SignTransaction`, `SignPersonalMessage`, and `SignAndExecuteTransaction` are using Zustand-backed auth/device/network hooks rather than React Query, and I didn’t find toast usage in those components either. If there is a specific missing context dependency behind the GraphQL/runtime error, can we point to that directly? Otherwise this looks broader than necessary and makes the real failure mode harder to reason about.

PR #78, opened 2026-03-25
fix: refresh expiry from oidc session

fcsondheim:
- refresh manual refreshjwt
- populate from oidc usermanager on init
- fix stale expiry


ccp-bofai (changes requested):
Probably best to read the comments. I am a little worried that in the future that certain functions could be called and produce incorrect results as they are meant to only be called when the data is fresh (at least that's how it looks to me). I believe the fix to this is simply tracking when the data was received and using that timestamp when needed so that the functions are stable.

ccp-bofai (changes requested):
I thought you were going to use `iat`? Also something else that I noticed which is basically "now is not the same as in one hour from now".

ccp-raudur:
1. Findings

`P1` `packages/shared/src/auth/storageService.ts:79-113`
The new composite JWT format has no migration path from the existing `Record<SuiChain, JwtResponse>` payload already stored under `evevault:jwt`. `normalizeJwtEntry()` only reads `candidate.primary` / `candidate.zkLogin`, so any legacy entry is converted to `{}` and effectively dropped. After this lands, `getJwtForNetwork()`, `hasJwtForNetwork()`, network rollback, and extension/web startup will all treat previously logged-in users as logged out until they re-authenticate. That is a release blocker for an auth-storage refactor.

`P2` `packages/shared/src/auth/stores/authStore.ts:117-125` and `apps/extension/src/lib/background/handlers/auth/dappLogin.ts:259-267`
Extension startup now refuses to hydrate from the stored network JWT unless `oidc-client-ts` already has a `User` persisted. But the background dapp login flow still only calls `storeJwt()`; it never mirrors that login into `UserManager`. Result: a user can successfully authenticate through a dapp, then open the popup and immediately appear logged out because `initialize()` exits at “no OIDC user in UserManager”. The previous implementation explicitly rebuilt the `User` from stored JWT for this case, so this PR regresses a working extension flow.

1b. Delta summary (new / resolved / remaining)

New:
- Composite JWT storage format without legacy-data migration.
- Extension init now depends on `UserManager` state that the background login flow does not populate.

Resolved:
- I did not find a new correctness issue in the `expires_at` / `expires_in` normalization itself.

Remaining:
- Both findings above still apply from the current PR head `f97c86b`.

2. Agent consensus

The auth refactor is directionally reasonable, but these two integration gaps make it unsafe to merge as-is. The storage-shape change needs backward compatibility, and the extension needs one source of truth consistently populated by every login entrypoint.

3. Agent disagreements

None.

4. Open questions or assumptions

I reviewed from a locally fetched PR branch because GitHub MCP access to `evefrontier/evevault` is blocked by org SAML on the available token. I also could not verify existing review comments for the same reason, so I may be repeating something already raised upstream.

5. Recommendation

Request changes. I would not merge until:
- `storageService` can read legacy `JwtResponse` entries and/or migrates them in place.
- The extension dapp-login path either stores the OIDC `User` too, or `initialize()` restores from persisted JWT when `UserManager` is empty.

fcsondheim:
P1 is safe to ignore as the expected pattern is a fresh install for every user and ignoring migration UX. Since this is still in heavy development it is irrelevant until we publish and we can expect updating through the store.

ccp-raudur (approved):
LGTM

packages/shared/src/auth/utils/__tests__/authStoreUtils.test.ts
```diff
@@ -49,10 +49,15 @@ describe("resolveExpiresAt", () => {
     vi.useRealTimers();
   });
 
+  // JWT with iat=1748779000 but no exp
+  const tokenWithIat =
+    "eyJhbGciOiJub25lIn0.eyJpYXQiOjE3NDg3NzkwMDAsInN1YiI6InUxIn0.";
```
ccp-bofai: I know this can be thrown into something online like jwt.io to see what is in there but this could have been manually constructed so there would be no need to trust the above comment about `iat` and `exp` (having the comment is good for emphasis though)

packages/shared/src/auth/storageService.ts
```diff
@@ -197,24 +196,24 @@ export async function clearZkLoginJwtForNetwork(
 }
 
 /**
- * Store a JWT for a specific network
+ * Store primary (OAuth) JWT for a specific network
  */
 export async function storeJwt(
```
ccp-bofai, line 201: The comment and the function name do not quite match. The function name suggests that it is generic in the type of JWT that is accepted. However the parameters suggest otherwise and this is where it gets confusing (auth0 docs point out: [ID Tokens](https://auth0.com/docs/secure/tokens/id-tokens) and [Access Tokens](https://auth0.com/docs/secure/tokens/access-tokens) and [Refresh Tokens](https://auth0.com/docs/secure/tokens/refresh-tokens) and also legacy Delegation Tokens. I guess this function is probably storing an access token or an ID token, perhaps the presence of the Nonce claim is important here too (mentioning if it is present or not, in name or comment, as that seems to be needed when talking to Enoki for example, so having the presence or absence of that made clear is probably helpful from a readability point of view)

ccp-bofai (commented):
Looks good to me, but there's still a few cases of "if I just look at this bit of a file, I am not sure if this is am ID Token or an Access Token or more to the point... do I have a token with a Nonce claim or not..." which I think would make it easier to see that we are handing an acceptable token over to Enoki when needed.

ccp-bofai (approved):
So approved :)

PR #84, opened 2026-04-10
fix: first login requires nonce for zklogin

fcsondheim:
include device nonce in initial jwt for Enoki during initial login at `/zklogin` endpoint

### DIFF
--- a/apps/extension/src/lib/background/services/oauthService.ts
+++ b/apps/extension/src/lib/background/services/oauthService.ts
@@ -14,8 +14,7 @@ function getAuthUrl(params: { tenantId: TenantId; nonce: string }) {
   url.searchParams.set("client_id", clientId);
   url.searchParams.set("redirect_uri", redirectUri);
   url.searchParams.set("scope", "openid profile email offline_access");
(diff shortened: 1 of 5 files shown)


vayan (approved):

apps/extension/src/lib/background/services/oauthService.ts
```diff
@@ -14,6 +14,10 @@ function getAuthUrl(params: { tenantId: TenantId }) {
   url.searchParams.set("client_id", clientId);
   url.searchParams.set("redirect_uri", redirectUri);
   url.searchParams.set("scope", "openid profile email offline_access");
+  // Always include the caller-provided nonce (zkLogin-derived from `initializeForChain`).
```
booninjah, line 17: If it should always be included, should it return an error if missing?

fcsondheim: That's a good shout, although in this instance the nonce is used only once through the lifecycle of the whole account-  i.e. at initialisation, when deriving account salt from the Enoki servers at the first time. Therefore nonce only needs to be included at this point once when the user is logging in for the first time, on subsequent logins it's safe even to omit this field entirely.

PR #88, opened 2026-04-15
feat: rotate ephkey on zkproof exp

fcsondheim:
- when zkproof expires, rotate ephkey to mitigate ephkey leak
 - Ephemeral key rotation on zkProof expiry: When a nonce is missing or the epoch is expired before generating a zkProof, the ephemeral key is now fully rotated (new keypair generated) instead of just re-initializing the chain.
    - All derived state (zkLogin JWTs, cached proofs) is cleared as part of the rotation.

- rotateEphemeralKey action: New deviceStore action that generates a fresh ephemeral keypair, clears all zkLogin JWTs and proofs, resets network data, and re-initializes the chain, keeping the session unlocked throughout.

- JWT persistence moved to session storage: JWTs are no longer written to disk-backed localStorage.
    - On extension, `chrome.storage.session` is used
    - On web, JWT ownership is delegated entirely to `oidc-client-ts` via a `sessionStorage`-backed `userStore`.

- On web, authStore init checks if the token is expired and attempts a silent renew via signinSilent() before hydrating the user, rather than falling through silently.


packages/shared/src/auth/storageService.ts
```diff
@@ -80,7 +47,7 @@ export async function getStoredChain(): Promise<SuiChain> {
 
 async function readJwtStorage(): Promise<JwtStorage | null> {
   if (isExtension()) {
-    const result = await chrome.storage.local.get([JWT_STORAGE_KEY]);
+    const result = await chrome.storage.session.get([JWT_STORAGE_KEY]);
     const raw = result[JWT_STORAGE_KEY];
     if (raw == null || typeof raw !== "object") return null;
```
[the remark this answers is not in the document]
fcsondheim: Destructive changes are permissible until we move to v0.1.0. Ignore

packages/shared/src/components/Layout/Header/HeaderMobile.tsx
```diff
@@ -36,22 +37,16 @@ export const HeaderMobile: React.FC<HeaderMobileProps> = ({
       items.push({
         label: "Dev mode",
         icon: "Settings" as IconName,
-        onClick: onDevModeToggle,
+        onClick: () => {},
```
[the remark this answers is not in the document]
fcsondheim: This is exactly the behaviour we are intending, that only clicking the Switch should trigger an interactive change.

packages/shared/src/auth/storageService.ts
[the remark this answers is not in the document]
fcsondheim: falling back to `chrome.storage.local` undermines the whole reason for the move to sessionStorage. Better to return a clear error instead.

apps/extension/entrypoints/keeper/keeper.ts
```diff
@@ -165,6 +198,45 @@ chrome.runtime.onMessage.addListener(
...
+          sendResponse({
+            ok: false,
+            error: error instanceof Error ? error.message : "Unknown error",
+          });
+        }
+      })();
+
+      return true;
+    }
+
```
booninjah, line 241: It would be great to get this in a seperate function

booninjah (approved):

packages/shared/src/auth/stores/authStore.ts
```diff
@@ -176,7 +176,41 @@ export const useAuthStore = create<AuthState>()(
...
+                log.info(
+                  "Web init: no session or refresh token, not logged in",
+                  {
+                    network,
+                  },
+                );
+                return set({ user: null, loading: false });
+              }
+              try {
+                webUser = await webUserManager.signinSilent();
```
ccp-raudur, line 197: P2: The new web init path stores the raw result of `signinSilent()` directly into auth state. That refreshed OIDC user will not include the locally-added `sui_address`/`salt` fields from the callback flow, but the wallet UI still depends on `user.profile.sui_address` for balances and send flows. After the first silent renew, the web wallet can appear authenticated while address-driven functionality breaks. This path should mirror the extension path and run `enrichUserWithZkLoginIfNeeded` before setting/storing the refreshed user.

apps/extension/entrypoints/keeper/keeper.ts
```diff
@@ -165,6 +198,45 @@ chrome.runtime.onMessage.addListener(
...
+        sendResponse({
+          ok: false,
+          error: "Vault must be unlocked again before rotating keypair",
+        });
+        return false;
+      }
+
+      (async () => {
+        try {
+          ephemeralKey = Ed25519Keypair.generate();
```
ccp-raudur: P2: `ROTATE_KEYPAIR` assigns a new `ephemeralKey` before `encryptWithKey()` has produced the replacement ciphertext. If re-encryption fails, the handler returns `ok: false`, but the keeper is already holding the new in-memory key while the persisted secret key and popup state still point at the old one. That split-brain state can invalidate subsequent proof/signature flows until the vault is re-locked. Generate/encrypt into locals first, then swap `ephemeralKey` only after persistence succeeds.

packages/shared/src/services/webVaultService.ts
```diff
@@ -162,6 +161,22 @@ class WebVaultService {
...
   }
 
+  async rotateEphemeralKeyPair(): Promise<PublicKey> {
+    if (!this.isUnlocked()) {
+      throw new Error("Vault must be unlocked again before rotating keypair");
+    }
+
+    // Generate a new keypair. The PIN hash in IndexedDB is unchanged — the
+    // user's PIN hasn't changed, only the ephemeral key has been rotated.
+    this.signer = await WebCryptoSigner.generate();
```
ccp-raudur: P2: The web rotation path replaces `this.signer` before the new exported keypair has been persisted. If the IndexedDB write fails, the current session keeps signing with the new in-memory key while storage still holds the old one. That creates the same live-vs-persisted key mismatch as the extension path and can break later operations in subtle ways. Keep the new signer in a local variable and commit it only after the IndexedDB write succeeds.

ccp-raudur (commented):
1. Findings
- [P2] Web silent renew drops zkLogin address enrichment.
- [P2] Rotate handler commits live key before persistence.
- [P2] Web rotation leaves signer/storage out of sync on write failure.

2. Agent consensus
- The overall direction is sound.
- The remaining issues are correctness problems in the new renewal and rotation flows.

3. Agent disagreements
- None.

4. Open questions or assumptions
- Static review only; I did not run tests or builds.

5. Recommendation
- Request changes before merge and make the rotation steps transactional before swapping the live key/signer.

ccp-bofai (approved):
LGTM... 2 things you might want to consider...
Argon2id instead of PBKDF2 but maybe that's not available in browsers yet. WASM might be an alternative though.
I am noticing the use of "KeyPair" where I think (if we want to be more consistently matching ML's naming) it should be "Keypair". Feel free to argue over this with me, but I was doing the same :)

apps/extension/entrypoints/keeper/keeper.ts
fcsondheim: good call! we currently have all keeper functions in the same file but we can split them up for readability

PR #89, opened 2026-04-16
feat: attestations checksums

fcsondheim:
Building off Scetrov's PR #49 to introduce checksums and attestations to Vault builds.


ccp-bofai (approved):
Looks good to me

PR #93, opened 2026-04-27
feat(localnet): backend plumbing

### LINKED ISSUE #59: Request to implement a localnet private key option

fcsondheim:
keypair pipeline, network store, signing primitive

   - Add localnet network type, isLocalnetChain helper, message/storage key constants
   - Extend SUI client to accept URL override; add gRPC epoch fetching for localnet
   - Implement Ed25519 keypair load/persist/restore pipeline:
     keeper (local.ts) → localVaultHandlers → chrome.storage → offscreenService restore on READY
   - Add rawSign: plain Ed25519 signing via keeper, returns { bytes, signature }
   - Extend networkStore with localnetUrl and localnet-aware setChain
   - Skip zkLogin vault init on localnet in deviceStore initActions/proofActions

### DIFF
--- a/apps/extension/entrypoints/keeper/local.ts
+++ b/apps/extension/entrypoints/keeper/local.ts
@@ -1,0 +1,107 @@
+import { ephSign } from "@evevault/shared";
+import type { IntentScope } from "@mysten/sui/cryptography";
+import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
+import type { BackgroundMessage } from "../../src/lib/background/types";
+
+type LocalnetState = {
+  localnetKey: Ed25519Keypair | null;
+};
+
+const localnetSetKeypair = async (
+  state: LocalnetState,
+  message: BackgroundMessage,
+  sendResponse: (response: {
+    ok: boolean;
+    address?: string;
+    error?: string;
+  }) => void,
+) => {
+  try {
+    const { privateKey } = message as { privateKey?: string };
+
+    if (!privateKey) {
+      sendResponse({ ok: false, error: "privateKey required" });
+      return;
+    }
+
+    // suiprivkey1... (Bech32): SDK handles this directly
+    if (privateKey.startsWith("suiprivkey")) {
+      state.localnetKey = Ed25519Keypair.fromSecretKey(privateKey);
+    } else {
+      throw new Error("Invalid private key");
+    }
+
+    sendResponse({
+      ok: true,
+      address: state.localnetKey.getPublicKey().toSuiAddress(),
+    });
+  } catch (error) {
+    sendResponse({
+      ok: false,
+      error: error instanceof Error ? error.message : "Unknown error",
+    });
+  }
+};
+
+const localnetGetAddress = (
+  state: LocalnetState,
+  sendResponse: (response: {
+    ok: boolean;
+    address?: string | null;
+    error?: string;
+  }) => void,
+) => {
+  if (!state.localnetKey) {
+    sendResponse({ ok: true, address: null });
+  } else {
+    sendResponse({
+      ok: true,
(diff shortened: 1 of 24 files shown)


apps/extension/src/lib/background/handlers/localVaultHandlers.ts
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      const response = await sendToKeeper({
+        type: KeeperMessageTypes.LOCALNET_SET_KEYPAIR,
+        privateKey: message.privateKey,
+      });
+      if (response?.ok) {
+        // Persist raw key in background storage so it survives keeper restarts
+        chrome.storage.local.set({
+          [LOCALNET_KEY_STORAGE_KEY]: message.privateKey ?? "",
+        });
```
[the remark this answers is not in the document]
fcsondheim: This is acceptable for now, as the privatekey is only used on localnet. Will be encrypted later.

packages/shared/src/types/networks.ts
```diff
@@ -19,6 +30,24 @@ export const AVAILABLE_NETWORKS: NetworkOption[] = [
...
+  isExt: boolean,
+): NetworkOption[] {
+  if (devMode && isExt) {
+    return [
+      ...AVAILABLE_NETWORKS,
+      { chain: SUI_LOCALNET_CHAIN, label: "Localnet", shortLabel: "LOCAL" },
+    ];
+  }
+  return AVAILABLE_NETWORKS;
+}
```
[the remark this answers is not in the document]
fcsondheim: will be implemented with #94 and #95

apps/extension/entrypoints/keeper/local.ts
```diff
@@ -0,0 +1,107 @@
...
+    const { privateKey } = message as { privateKey?: string };
+
+    if (!privateKey) {
+      sendResponse({ ok: false, error: "privateKey required" });
+      return;
+    }
+
+    // suiprivkey1... (Bech32): SDK handles this directly
+    if (privateKey.startsWith("suiprivkey")) {
+      state.localnetKey = Ed25519Keypair.fromSecretKey(privateKey);
```
ccp-bofai, line 29: What is going to happen if the key is of scheme Secp256r1? Note that also `Secp256r1Keypair.fromSecretKey` accepts a Bech32 secret key string. You could use `sui.cryptography.decodeSuiPrivateKey(value: string): { scheme: SignatureScheme, secretKey Uint8Array}` to work out what you have if you want to support multiple keypair types.

apps/extension/entrypoints/keeper/local.ts
```diff
@@ -0,0 +1,107 @@
+import { ephSign } from "@evevault/shared";
+import type { IntentScope } from "@mysten/sui/cryptography";
+import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
+import type { BackgroundMessage } from "../../src/lib/background/types";
+
+type LocalnetState = {
+  localnetKey: Ed25519Keypair | null;
```
ccp-bofai, line 7: See comment below about `fromSecretKey`, if that makes sense then you might want to change this to just `Keypair` possibly.

ccp-bofai (approved):
If this is Ed25519 only then I think it looks good. If not then you might want to check the comments I added (but I did see a comment after writing them further down that this was Ed25519 only)

apps/extension/entrypoints/keeper/local.ts
fcsondheim: The extension is set up to support only Ed25519 at the moment, both as an ephemeral private key in the zklogin flow as well as now with the local-only key. If we want to support multiple keypair types that would be a different PR I think to add Secp256r1 across both instances.

PR #94, opened 2026-04-27
feat(localnet): wallet hook layer

### LINKED ISSUE #59: Request to implement a localnet private key option

fcsondheim:
balance, signing context, unified auth                           
                                                                                                        
   - Add gRPC balance fetching path in useBalance for localnet (no GraphQL endpoint)                    
   - Add signForChain: routes to rawSign (Ed25519) on localnet, zkSignAny on all others                 
   - Add useWalletSigningContext: single hook providing senderAddress, suiClient,
     isAuthenticated, isWalletUnlocked, and bound sign() for both signing paths
   - Add useLocalnetAddress, useActiveSuiAddress (delegates to context)
   - Add useIsAuthenticated (delegates to context; callers need no network awareness)
   - Refactor useSendToken and useDevMode to use sign() from context;
     remove per-call threading of chain/user/localnetAddress/getZkProof
   - Thread balanceAddress + localnetUrl through TokenSection for localnet balance

### DIFF
--- a/packages/shared/src/wallet/hooks/__tests__/useBalance.test.tsx
+++ b/packages/shared/src/wallet/hooks/__tests__/useBalance.test.tsx
@@ -1,4 +1,4 @@
-import { SUI_DEVNET_CHAIN } from "@mysten/wallet-standard";
+import { SUI_DEVNET_CHAIN, SUI_LOCALNET_CHAIN } from "@mysten/wallet-standard";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { renderHook, waitFor } from "@testing-library/react";
 
@@ -6,9 +6,14 @@ import type { ReactNode } from "react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 const mockQuery = vi.fn();
+const mockGetBalance = vi.fn();
 
 vi.mock("../../../sui/graphqlClient", () => ({
   createSuiGraphQLClient: vi.fn(() => ({ query: mockQuery })),
+}));
+
+vi.mock("@evevault/shared/sui", () => ({
+  createSuiClient: vi.fn(() => ({ getBalance: mockGetBalance })),
 }));
 
 vi.mock("@evevault/shared/utils", () => ({
@@ -22,7 +27,9 @@ vi.mock("@evevault/shared/utils", () => ({
   isWeb: vi.fn(() => true),
   isBrowser: vi.fn(() => true),
   SUI_COIN_TYPE: "0x2::sui::SUI",
-  formatByDecimals: vi.fn((balance: string) => balance),
+  formatByDecimals: vi.fn(
+    (balance: string, _decimals: number) => `formatted-${balance}`,
+  ),
   formatMistToSui: vi.fn(),
 }));
 
@@ -41,6 +48,7 @@ const createWrapper = (queryClient: QueryClient) => {
 describe("useBalance hook", () => {
   beforeEach(() => {
     mockQuery.mockClear();
+    mockGetBalance.mockClear();
   });
 
   it("returns a formatted SUI balance for the current user", async () => {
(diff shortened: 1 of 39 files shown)


packages/shared/src/wallet/hooks/__tests__/useBalance.test.tsx
```diff
@@ -171,3 +179,114 @@ describe("useBalance hook", () => {
...
+    expect(result.current.data?.formattedBalance).toBe("2");
+    expect(result.current.data?.rawBalance).toBe("2000000000");
+    unmount();
+    queryClient.clear();
+  });
+
+  it("uses 9-decimal fallback and warns for unknown localnet tokens", async () => {
+    mockGetBalance.mockResolvedValue({ balance: { balance: "5000000000" } });
+    const { createLogger } = await import("@evevault/shared/utils");
+    const mockWarn = vi.fn();
```
ccp-bofai: The description of this test suggests that it "warns for unknown localness tokens". I am not seeing how that is being done but it looks like `mockWarn` might have been part of this. So I am not sure if code is missing or the description is not quite right?

packages/shared/src/wallet/hooks/__tests__/useSendToken.test.tsx
```diff
@@ -78,15 +86,16 @@ vi.mock("@mysten/sui/transactions", () => {
 // Import after mocks
 // Using workspace aliases in test files due to Vite resolution limitations with relative imports
 import { getUserForNetwork, useAuth } from "@evevault/shared/auth";
-import { useDevice } from "@evevault/shared/hooks";
 import { useNetworkStore } from "@evevault/shared/stores/networkStore";
 import { createSuiClient } from "@evevault/shared/sui";
 import { createMockUser } from "@evevault/shared/testing";
+import { useDevice } from "../../../hooks/useDevice";
```
ccp-bofai, line 92: I have dealt with this in another project by adding something like the following to `defineConfig` in the `vitest.config.ts` file:
```
  resolve: {
    alias: [
      {
        find: /^#src\//,
        replacement: path.resolve(__dirname, 'src') + '/',
      },
      {
        find: /^#tests\//,
        replacement: path.resolve(__dirname, 'tests') + '/',
      },
    ],
  },
```
and before that I also added the following so `__dirname` was defined:
```
import path from 'node:path'
import { fileURLToPath } from 'node:url'

const __dirname = path.dirname(fileURLToPath(import.meta.url))
```

packages/shared/src/wallet/hooks/useBalance.ts
```diff
@@ -1,7 +1,9 @@
-import { SUI_TESTNET_CHAIN } from "@mysten/wallet-standard";
+import { SUI_LOCALNET_CHAIN, SUI_TESTNET_CHAIN } from "@mysten/wallet-standard";
 import { useQuery } from "@tanstack/react-query";
 import { useMemo } from "react";
+import { createSuiClient } from "../../sui";
```
ccp-bofai, line 4: In `package.json` you might want to consider something like I use elsewhere:
```
  "imports": {
    "#src/*": [
      "./src/*",
      "./src/*.ts",
      "./src/*/index.ts"
    ],
    "#tests/*": [
      "./tests/*",
      "./tests/*.ts",
      "./tests/*/index.ts"
    ]
  },
```
So imports like that are from '#src/...' or '#tests/...' (tests are not under `#src/` where I do this but I guess there is some selection mechanism to not ship `__tests__` directories)

packages/shared/src/wallet/hooks/useLocalnetAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { chain } = useNetworkStore();
+  const [localnetAddress, setLocalnetAddress] = useState<string | null>(null);
+
+  useEffect(() => {
+    let isCancelled = false;
+
+    if (isLocalnetChain(chain) && isExtension()) {
+      localnetKeyService
+        .getAddress()
+        .then((addr) => {
```
ccp-bofai: Minor niggle `.then` while I see async/await elsewhere. React also has an example of how to use async/await in [useEffect](https://react.dev/reference/react/useEffect#fetching-data-with-effects). Maybe something for a different PR to move over to async/await a bit more

ccp-bofai (changes requested):
Looks good but had some comments on things that you might want to make note of for improvements in later PRs.
Requesting change as there is one test that I am not understanding how it achieves its intention (the textual description does not quite add up with the code in the test as far as I can tell)

packages/shared/src/wallet/hooks/__tests__/useSendToken.test.tsx
fcsondheim: Nice! This is something we'll definitely add to a next PR for more general code cleanup

packages/shared/src/wallet/hooks/useBalance.ts
fcsondheim: same as above, will add in a code cleanup PR!

ccp-bofai (approved):
Looks good to me

PR #95, opened 2026-04-27
featl(localnet) - ui changes

### LINKED ISSUE #59: Request to implement a localnet private key option

fcsondheim:


apps/extension/src/routes/localnet-settings.tsx
```diff
@@ -0,0 +1,200 @@
...
+  }, []);
+
+  const handleUrlSave = useCallback(async () => {
+    const trimmed = urlDraft.trim();
+    if (!trimmed) return;
+    setLocalnetUrl(trimmed);
+    setUrlStatus("loading");
+    try {
+      await useDeviceStore.getState().initializeForChain(SUI_LOCALNET_CHAIN);
+      setUrlStatus("ok");
```
ccp-raudur, line 78: P2: This can report success for an unreachable localnet URL. `initializeForChain(SUI_LOCALNET_CHAIN)` currently swallows RPC/epoch fetch failures in the localnet init path, so this line can set `Connected` after persisting a bad URL. Please validate the RPC URL with a call that rejects on failure, then persist/show success only after that succeeds.

apps/extension/src/routes/send-token.tsx
```diff
@@ -33,7 +40,13 @@ function SendTokenPage() {
 }
 
 export const Route = createFileRoute("/send-token")({
-  beforeLoad: () => requireAuth(),
+  beforeLoad: () => {
```
ccp-raudur: P2: This guard allows unauthenticated localnet access, but it does not verify that a localnet keypair/address is loaded. Direct navigation to `/send-token` in that state renders a non-functional wallet path and can pass undefined values into the header/send flow. Please redirect localnet users without an address to `/localnet-settings` before rendering this route.

apps/extension/src/routes/localnet-settings.tsx
```diff
@@ -0,0 +1,200 @@
...
+          </Text>
+
+          <Input
+            label="Private key"
+            onChange={(e) => {
+              setPrivateKeyDraft(e.target.value);
+              setKeyStatus("idle");
+              setKeyError(null);
+            }}
+            onKeyDown={(e) => e.key === "Enter" && void handleUrlSave()}
```
ccp-raudur: P3: Pressing Enter in the private-key field currently saves the RPC URL instead of loading the keypair. This should call `handleKeySave()` so keyboard submission matches the Load Key button.

packages/shared/src/components/NetworkSelector/NetworkSelector.tsx
```diff
@@ -42,22 +52,29 @@ export const NetworkSelector: React.FC<NetworkSelectorProps> = ({
       } else if (result.requiresReauth) {
         onNetworkSwitchStart?.(chain, targetChain);
         onRequiresReauth?.(targetChain);
+      } else if (targetChain === SUI_LOCALNET_CHAIN && isExtensionContext) {
+        await onLocalnetSelected?.();
       }
 
       setIsProcessing(false);
     },
```
ccp-raudur, line 73: P3: If `setChain()` or `onLocalnetSelected()` rejects, execution skips this reset and leaves the selector disabled until reload. Please wrap the async switch in `try/finally` so processing state is always cleared.

apps/extension/src/routes/localnet-settings.tsx
```diff
@@ -0,0 +1,233 @@
...
+  >("idle");
+  const [keyError, setKeyError] = useState<string | null>(null);
+
+  const [address, setAddress] = useState<string | null>(null);
+
+  useEffect(() => {
+    localnetKeyService
+      .getAddress()
+      .then(setAddress)
+      .catch(() => setAddress(null));
```
ccp-bofai: Looks a bit out of place when there is async/await code further down in the same file

fcsondheim: old habits die hard! will fix

ccp-bofai (approved):
Looks good to me. Just one comment about a `.then().catch()` where there is also async/await code nearby. Something that could be addressed in another PR

PR #96, opened 2026-04-28
chore: code cleanup

fcsondheim:
avoid relative imports.

self-imports in `shared` now read from `#/`
self-imports in `apps/extension` and `apps/web` now read from `@/`

remove excessive agent guidance


ccp-bofai (approved):
Looks good to me

PR #97, opened 2026-04-29
feat: encrypt localnet key

fcsondheim:
**Localnet key encryption (PIN-protected)**
- Localnet private keys are now encrypted with the user's PIN (AES-GCM, same scheme as the zkLogin ephemeral key) before being stored in `chrome.storage.local` — previously stored as plain text
- Locking the wallet now clears the localnet keypair from keeper RAM; unlocking automatically restores it by decrypting the stored blob with the PIN
- `useLocalnetAddress` now clears immediately when the vault locks, and `isWalletUnlocked` requires `!isLocked` for localnet the same as zkLogin

**Authentication parity between localnet and zkLogin**
- Localnet no longer bypasses OIDC authentication — both modes require a valid user session and an unlocked vault
- Lock screen now shows whenever `isLocked` is true
- After logging out, FusionAuth sign-in is also required to access the wallet

**Code cleanup**
- Removed `networkUser` async state from `useWalletSigningContext`; `useAuth().user` is used directly as the source of truth
- Removed dead code: `setPreviousNetworkBeforeSwitch` (unimplemented rollback feature in `PopupApp`), `useIsAuthenticated` hook (no callers), and the unencrypted localnet key restore-on-keeper-restart path in `offscreenService`
- Fixed `TransactionsScreen` Biome `useSemanticElements` violation by replacing `role="button"` divs with real `<button>` elements
- Added recovery path on `LockScreen` for new users who hit a keypair initialization error (`deviceError` now also shows the reset option)


apps/extension/src/lib/background/handlers/vaultHandlers.ts
```diff
@@ -71,10 +75,21 @@ export async function handleUnlockVault(
...
   try {
+    // Read the encrypted localnet key (if any) so keeper can restore it during unlock
+    const stored = await chrome.storage.local.get(LOCALNET_STORAGE_KEY);
+    const storedLocalnet = stored[LOCALNET_STORAGE_KEY];
+    const encryptedLocalnetKey =
+      storedLocalnet &&
+      typeof storedLocalnet === "object" &&
+      "data" in storedLocalnet
+        ? storedLocalnet
+        : null;
```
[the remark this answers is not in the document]
fcsondheim: safe to ignore; the legacy plaintext value is not in a released version so should not be relevant to any users

apps/extension/entrypoints/keeper/__tests__/keeper.localnet.test.ts
```diff
@@ -0,0 +1,234 @@
...
+    } catch (error) {
+      sendResponse({
+        ok: false,
+        error: error instanceof Error ? error.message : "Unknown error",
+      });
+    }
+  }
+
+  it("loads keypair into RAM and returns encrypted blob when vault is unlocked", async () => {
+    const { bech32 } = makeTestKey();
```
ccp-bofai, line 90: I think this line should be moved down in the test to where it is used i.e. just before the line `await simulateSetKeypair(bech32);`. So its usage is clearer given the locality, which it currently does not have.
This pattern is present in other tests in this file.

apps/extension/entrypoints/keeper/__tests__/keeper.localnet.test.ts
```diff
@@ -0,0 +1,234 @@
...
+    const { encryptedKey } = sendResponse.mock.calls[0][0] as {
+      encryptedKey: HashedData;
+    };
+    const decrypted = await decrypt(encryptedKey, TEST_PIN);
+    expect(decrypted).toBe(bech32);
+  });
+
+  it("returns error when vault is locked (no session key)", async () => {
+    const { bech32 } = makeTestKey();
+
```
ccp-bofai, line 127: Might be an idea to add a comment to explain the "no session key" in the test description. e.g. something like `// We are not calling makeSessionKey so there is no session key`. Reading the test as is this is not obvious without reading other similar tests to establish the difference.

apps/extension/entrypoints/keeper/__tests__/keeper.localnet.test.ts
```diff
@@ -0,0 +1,234 @@
...
+      error: "Vault must be unlocked to store localnet key",
+    });
+  });
+
+  it("returns error when privateKey is missing", async () => {
+    const hashedKey = await encrypt(makeTestKey().bech32, TEST_PIN);
+    sessionDerivedKey = await makeSessionKey(TEST_PIN, hashedKey.salt);
+    sessionSalt = hashedKey.salt;
+
+    await simulateSetKeypair(undefined);
```
ccp-bofai, line 141: A comment above this might help reviewers establish that this is the "when privateKey is missing" bit from the tests title. i.e. `// The missing private key is passed as 'undefined' to `simulateSetKeypair`

apps/extension/entrypoints/keeper/__tests__/keeper.localnet.test.ts
```diff
@@ -0,0 +1,234 @@
...
+
+    expect(localnetKey).toBeNull();
+    expect(sendResponse).toHaveBeenCalledWith({
+      ok: false,
+      error: "privateKey required",
+    });
+  });
+
+  it("returns error when private key does not start with suiprivkey", async () => {
```
ccp-bofai, line 150: 'suiprivkey' probably should be 'suiprivkey1' here I believe

apps/extension/entrypoints/keeper/local.ts
```diff
@@ -8,33 +8,38 @@ type LocalnetState = {
...
-    if (!privateKey) {
-      sendResponse({ ok: false, error: "privateKey required" });
-      return;
-    }
-
-    // suiprivkey1... (Bech32): SDK handles this directly
-    if (privateKey.startsWith("suiprivkey")) {
-      state.localnetKey = Ed25519Keypair.fromSecretKey(privateKey);
-    } else {
+    if (!privateKey.startsWith("suiprivkey")) {
```
ccp-bofai: Might want to use `SUI_PRIVATE_KEY_PREFIX` which is exported from `@mysten/sui/cryptography` rather than "suiprivkey" possibly.

apps/extension/src/lib/background/handlers/__tests__/vaultHandlers.localnet.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    hashedSecretKey: HASHED_SECRET_KEY,
+    pin: "123456",
+    ...overrides,
+  } as unknown as VaultMessage;
+}
+
+function installChromeMock(
+  storedValue: unknown,
+  keeperResponse: unknown = { ok: true },
+) {
```
ccp-bofai: When reading the tests where this is used I get the feeling that the function name does not help in making the test more understandable (even though the function name does describe exactly what is being done). This is a bit of a predicament and I think that if the parameters were passed as named properties in an object then the intent would be more obvious, perhaps consider renaming "storedValue" to "localStorageValue" to make it blatantly obvious what is happening.

apps/extension/src/lib/background/handlers/__tests__/vaultHandlers.test.ts
```diff
@@ -0,0 +1,414 @@
...
+  return { ...actual, LOCALNET_STORAGE_KEY: "evevault:localnet-key" };
+});
+
+const mockSender = {} as chrome.runtime.MessageSender;
+
+function makeMessage(overrides: Partial<VaultMessage> = {}): VaultMessage {
+  return { type: "VAULT_MSG", ...overrides } as unknown as VaultMessage;
+}
+
+function installChromeMock(keeperResponse: unknown = { ok: true }) {
```
ccp-bofai: Same issue with intent/readability as the function with the same name in `vaultHandlers.localnet.test.ts`.

apps/extension/src/lib/background/handlers/__tests__/vaultHandlers.test.ts
```diff
@@ -0,0 +1,414 @@
...
+
+    _handleRotateKeypair(makeMessage(), mockSender, sendResponse);
+    await new Promise((r) => setTimeout(r, 0));
+
+    expect(sendResponse).toHaveBeenCalledWith({
+      ok: false,
+      error: "Vault must be unlocked again before rotating keypair",
+    });
+  });
+});
```
ccp-bofai, line 220: After reading both these tests I am wondering if the "fallback error" is `error: "Vault must be unlocked again before rotating keypair"` as suggested from the test titled "returns fallback error when keeper fails with no message". When I look at the test above it I am noticing that this appears to be setting and returning the same text. Thus it is difficult to gauge if the first test is returning the error text that was set or the fallback error text.

ccp-bofai (changes requested):
I like this but some of the tests could be a bit easier to understand with some small changes.

apps/extension/entrypoints/keeper/__tests__/keeper.localnet.test.ts
fcsondheim: The private key is only guaranteed to start with `suiprivkey` without `1`: https://sdk.mystenlabs.com/typedoc/functions/_mysten_sui.cryptography.decodeSuiPrivateKey.html

ccp-bofai: Its in Bech32 format so checking for the presence of the separator, which is the "1" character after the human readable part ensures that you have something that looks like its valid Bech32.

ccp-bofai (approved):
Looks good to me, but be aware of Bech32. It is kind of interesting, especially its own alphabet that avoids look alike characters and so on.

PR #98, opened 2026-04-29
fix(keeper): persist ephemeral key to chrome.storage.session to survive offscreen teardown

r4wf0d0g23:
## Problem

The keeper offscreen document at `apps/extension/entrypoints/keeper/keeper.ts` holds the unlocked ephemeral key (`ephemeralKey`, `_vaultUnlocked`, `_vaultUnlockExpiry`) in module-level RAM.

In Chromium MV3, the offscreen document is closed when the service worker is torn down — which happens after ~30 seconds of service-worker idle. When the offscreen is reconstructed, all module-level state is lost. The next signing request from a dApp falls into the `if (!ephemeralKey)` branch and returns `[KEEPER_EPH_SIGN] LOCKED`, even though the user's 10-minute unlock timer never expired.

End users perceive this as the vault "locking itself" between transactions and have to re-enter their PIN repeatedly. We see this in the wild on the CradleOS dApp (sui-testnet/Stillness) — a transaction submitted ~30 s after the previous one consistently fails with `LOCKED`.

## Fix

Persist `(ephemeralKey, unlockExpiry)` to `chrome.storage.session` whenever state changes, restore at offscreen module load, and have read paths await the in-flight restore so a cold-start race does not falsely report `LOCKED`.

### Why `chrome.storage.session`?

`chrome.storage.session` is **in-memory only, never written to disk**, and **isolated per extension**. It matches the existing "ephemeral key" security model exactly:

| Property | Module-level RAM (today) | `chrome.storage.session` (this PR) |
|---|---|---|
| Cleared on browser exit | ✅ | ✅ |
| Cleared on extension reload | ✅ | ✅ |
| Written to disk | ❌ | ❌ |
| Survives offscreen teardown | ❌ | ✅ |
| Visible to other extensions | ❌ | ❌ |

The user PIN is **never** persisted. WebCrypto-derived `sessionDerivedKey` / `sessionSalt` are intentionally **not** restored; key rotation will simply require a fresh unlock if the offscreen has been torn down. This preserves the existing security boundary for the rotation flow while fixing the much more common signing flow.

### Code changes

1. **New module `sessionPersistence.ts`** with three exported helpers:
   - `persistKeeperState(keypair, expiry)` — best-effort write
   - `restoreKeeperState()` — best-effort read with built-in expiry check
   - `clearPersistedKeeperState()` — best-effort delete
2. **`keeper.ts`**:
   - On module load, kicks off `restoreSessionStateIfAny()` and stores the promise.
   - Persists state in `CREATE_KEYPAIR`, `UNLOCK_VAULT`, and `ROTATE_KEYPAIR`.
   - Clears persisted state in `CLEAR_EPHKEY` and `checkAndEnforceExpiry`.
   - Read handlers (`GET_PUBLIC_KEY`, `EPH_SIGN`, `SET_ZKPROOF`, `GET_ZKPROOF`) now `await restorePromise` before checking the lock, so a cold-start signing race does not falsely return `LOCKED`.

### Manifest

The `storage` permission is **already declared** in `apps/extension/wxt.config.ts`. No manifest changes required.

## Tests

- 6 new vitest cases in `apps/extension/entrypoints/keeper/__tests__/sessionPersistence.test.ts` covering:
  - round-trip persist + restore
  - returns null when no state has been persisted
  - returns null and clears storage when persisted state has expired
  - explicit clear removes the entry
  - returns null gracefully when `chrome.storage.session` is unavailable
  - `persistKeeperState` swallows errors (best-effort)
- All 13 keeper tests pass (`bun run test:run entrypoints/keeper/__tests__/`):

```
 ✓ entrypoints/keeper/__tests__/sessionPersistence.test.ts (6 tests) 31ms
 ✓ entrypoints/keeper/__tests__/keeper.lock.test.ts (7 tests) 49ms

 Test Files  2 passed (2)
      Tests  13 passed (13)
```

- `bunx biome check` clean on all three touched files.

## Reproduction (before this PR)

1. Unlock EVE Vault.
2. Sign a transaction in any dApp. ✅
3. Wait 30+ seconds without interacting with the extension.
4. Sign another transaction. → `Error: [KEEPER_EPH_SIGN] LOCKED`

After this PR, step 4 succeeds without re-prompting for PIN, until the 10-minute unlock window expires naturally.

## Risks / non-goals

- **Browser compatibility**: `chrome.storage.session` is supported in Chromium 102+ (May 2022) and Firefox MV3. The extension already requires MV3.
- **Storage quota**: `chrome.storage.session` allows ~10 MB per extension; we store one ~80 byte entry. No quota concern.
- **Multiple offscreen instances**: Chrome enforces a single offscreen document per extension, so there is no concurrency hazard.
- **Out of scope for this PR**: persisting `zkProofs` or rotation-derived WebCrypto state. Both can be added incrementally if desired.

## Discord context

Reported by users in the EVE Frontier dev community. Happy to provide repro logs or tweak any of this — just leave a review comment.

### DIFF
--- a/apps/extension/entrypoints/keeper/__tests__/sessionPersistence.test.ts
+++ b/apps/extension/entrypoints/keeper/__tests__/sessionPersistence.test.ts
@@ -1,0 +1,113 @@
+/// <reference types="chrome"/>
+
+import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  clearPersistedKeeperState,
+  persistKeeperState,
+  restoreKeeperState,
+} from "../sessionPersistence";
+
+// Minimal in-memory mock of chrome.storage.session for vitest.
+function makeSessionStorageMock() {
+  const store = new Map<string, unknown>();
+  return {
+    async get(key?: string | string[]) {
+      if (typeof key === "string") {
+        return store.has(key) ? { [key]: store.get(key) } : {};
+      }
+      const result: Record<string, unknown> = {};
+      for (const [k, v] of store.entries()) result[k] = v;
+      return result;
+    },
+    async set(items: Record<string, unknown>) {
+      for (const [k, v] of Object.entries(items)) store.set(k, v);
+    },
+    async remove(key: string | string[]) {
+      const keys = Array.isArray(key) ? key : [key];
+      for (const k of keys) store.delete(k);
+    },
+    _store: store,
+  };
+}
+
+describe("keeper sessionPersistence", () => {
+  let mockSession: ReturnType<typeof makeSessionStorageMock>;
+
+  beforeEach(() => {
+    mockSession = makeSessionStorageMock();
+    (globalThis as unknown as { chrome: unknown }).chrome = {
+      storage: { session: mockSession },
+    };
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("round-trips an ephemeral keypair through persist + restore", async () => {
+    const original = Ed25519Keypair.generate();
+    const expiry = Date.now() + 5 * 60 * 1000;
+
+    await persistKeeperState(original, expiry);
+    const restored = await restoreKeeperState();
+
+    expect(restored).not.toBeNull();
+    expect(restored?.unlockExpiry).toBe(expiry);
+    expect(restored?.ephemeralKey.getPublicKey().toRawBytes()).toEqual(
+      original.getPublicKey().toRawBytes(),
+    );
+  });
+
+  it("returns null when no state has been persisted", async () => {
+    const restored = await restoreKeeperState();
+    expect(restored).toBeNull();
+  });
+
+  it("returns null and clears storage when persisted state has expired", async () => {
+    const original = Ed25519Keypair.generate();
+    const expiredAt = Date.now() - 1000;
+
+    await persistKeeperState(original, expiredAt);
+    const restored = await restoreKeeperState();
+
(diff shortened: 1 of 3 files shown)


fcsondheim:
The current RAM-only design with the ephemeral key in the keeper is intentional as it holds the key as an opaque JS object confined to the offscreen document's renderer. Moving this to `chrome.storage.session` would hold the key as a readable string that is accessible in any extension context. The key would be moved from the extension renderer into the browser's main process, which would introduce a new vulnerability and increases the attack surface.

The UX bug identified is accurate, but I think we can handle this differently. A better path would be to introduce a keepalive ping to the service worker and offscreen document while the vault is unlocked.

fcsondheim:
For the future, perhaps it might be good to start by opening this as an issue to come to a good proposed solution

fcsondheim:
@r4wf0d0g23 I was trying to recreate this bug with [v.0.0.9](https://github.com/evefrontier/evevault/releases/tag/v0.0.9) for #99 but wasn't able to when signing with some of the dapps I was using. Can you provide a link to the dapp you were trying this with when you first ran into this UX issue?

r4wf0d0g23:
@fcsondheim i have also not been able to reproduce this since v.0.0.9 and have also not seen any [KEEPER_EPH_SIGN]. I was originally seeing the errors in https://r4wf0d0g23.github.io/CradleOS/ when signing for verification after timeout, page refreshes, multiple signing loops, or when toggling assemblies on/off.

fcsondheim:
@r4wf0d0g23 That's great to hear, if that's case the UX issue identified both here and in #99 are no longer applicable. I'll close #99 for now but please let us know if the issue reoccurs.

PR #101, opened 2026-05-07
chore: migrate network and tenant state to context

fcsondheim:
migrate network, tenant stores to a unified context store


packages/shared/src/stores/contextStore.ts
```diff
@@ -0,0 +1,234 @@
...
+function getInitialStateFromLocalStorage(): PersistedContextState {
+  if (!isWeb() || typeof window === "undefined" || !window.localStorage) {
+    return {};
+  }
+
+  const context = parsePersistedState<PersistedContextState>(
+    window.localStorage.getItem(CONTEXT_STORAGE_KEY),
+  );
+  return context ?? {};
+}
```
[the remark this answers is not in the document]
fcsondheim: safe to ignore

packages/shared/src/utils/storageKeys.ts
```diff
@@ -1,21 +1,19 @@
...
 export const EVEVAULT_STORAGE_KEYS = [
   AUTH_STORAGE_KEY,
   DEVICE_STORAGE_KEY,
-  NETWORK_STORAGE_KEY,
+  CONTEXT_STORAGE_KEY,
   TOKENLIST_STORAGE_KEY,
   JWT_STORAGE_KEY,
-  TENANT_STORAGE_KEY,
   LOCALNET_STORAGE_KEY,
 ] as const;
```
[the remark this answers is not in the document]
fcsondheim: safe to ignore

packages/shared/src/auth/resetVaultOnDevice.ts
```diff
@@ -84,7 +86,10 @@ function resetStoresToInitial(): void {
   });
 
   useNetworkStore.setState({
+    tenantId: TenantId.STILLNESS,
+    devMode: false,
     chain: SUI_TESTNET_CHAIN,
```
ccp-bofai, line 91: I am still having some reservations about seeing an object that references a tenant and a chain in this hard coded manner. It is probably the risk of this being incorrect when the chain that is normally associated with the tenant changes that is the root of that. I think this is something to think about but perhaps having these values in the JWT is better so that when we do change, it just needs to be at that level rather than requiring a new version of Eve Vault to be published.

packages/shared/src/stores/contextStore.ts
```diff
@@ -0,0 +1,238 @@
...
+const initialState = getInitialStateFromLocalStorage();
+
+export const useContextStore = create<ContextState>()(
+  persist(
+    (set, get) => ({
+      tenantId: initialState.tenantId ?? INITIAL_TENANT_ID,
+      devMode: initialState.devMode ?? false,
+      setTenantId: async (id: TenantId) => {
+        if (!getAvailableTenantIds(true).includes(id)) {
+          return;
```
ccp-bofai, line 67: Would you not want to log this?

ccp-bofai (approved):
Looks good to me.

packages/shared/src/auth/resetVaultOnDevice.ts
fcsondheim: The hardcoding is only referenced in this one file which resets the state. Wouldn't it be expected to reset to the hardcoded/default values here?

ccp-bofai: Good to have defaults but I am not sure if this implies a relationship between Stillness and sui:testnet that is fixed (in code), if you see what I mean, because that will change at some point in time I believe (probably after some cycle and wipe).

fcsondheim: I don't think so. But even then, we don't have a central config really with this pairing I think; but it's a good call to have fewer of these types of values hardcoded.

PR #103, opened 2026-05-08
test: expand test coverage

fcsondheim:
- **Replaces simulated keeper tests** (`keeper.lock.test.ts`, `keeper.localnet.test.ts`) with integration tests (`keeper.vault.test.ts`, `keeper.operations.test.ts`) that import and exercise the real `keeper.ts` handler
- **Expands test coverage** across `authStore` (initialize, login, extensionLogin, tenant switch, hydration), routing utilities, `TenantUrlSync`, `useResponsive`, `useCopyToClipboard`, `useAuth`, SUI GraphQL client/epoch helpers, `webVaultService`, zkProof, and transaction utilities
- **CI**: uploads coverage summaries from `main` as a baseline artifact and fails PRs that drop coverage; moves tests from pre-commit to pre-push hook; adds web app coverage reporting

## Bug fixes (surfaced by tests)

- **`keeper.ts` / `webVaultService.ts`**: `ROTATE_KEYPAIR` was incorrectly extending the vault's 10-minute unlock expiry on every rotation — a successful rotation could keep the vault open indefinitely. Expiry now set only on unlock
- **`authStore` web login**: `initializeForChain` errors were uncaught, leaving `loading: true` permanently. Wrapped in try/catch; non-zkLogin chains now return early instead of silently proceeding
- **`runTenantSwitchCleanup`**: was locking the device vault during a tenant switch, which logged the user out of their active session unintentionally
- **`switchTenantAndReload`**: was redirecting to a constructed URL (losing query params, breaking tenant routing edge cases) instead of simply reloading
- **`useResponsive`**: crashed when `window.innerWidth` was `undefined` (some embedded environments); now falls back to 1200px
- **`deviceStore`**: `localnetUrl` was missing from initial state, causing the field to be `undefined` until explicitly set

## Test plan
- [x] `bun run test:run` passes (117 extension + 59 web tests + 451 shared tests)
- [x] Coverage baseline uploads on merge to `main`
- [ ] Coverage regression check runs on PRs that have a baseline available


apps/extension/entrypoints/keeper/__tests__/keeper.vault.test.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    vi.useFakeTimers({ toFake: ["Date"] });
+    vi.setSystemTime(Date.now() + 11 * 60 * 1000);
+
+    const resp = await dispatch({ type: KeeperMessageTypes.ROTATE_KEYPAIR });
+    expect(resp.ok).toBe(false);
+    expect(String(resp.error)).toContain("Vault must be unlocked again");
+  });
+
+  it("EPH_SIGN returns LOCKED when neither ephemeralKey nor localnetKey is set", async () => {
```
ccp-bofai: "neither" / "nor" usage is potentially confusing for people. Could this please be stated in simpler terms please.

packages/shared/src/auth/stores/__tests__/authStore.extensionLogin.test.ts
```diff
@@ -0,0 +1,271 @@
...
+type ChromeMessageListener = (message: {
+  id: string;
+  type: string;
+  token?: unknown;
+  error?: unknown;
+}) => void;
+
+function makeJwtPayload(claims: Record<string, unknown>): string {
+  const payload = Buffer.from(JSON.stringify(claims)).toString("base64url");
+  return `eyJhbGciOiJIUzI1NiJ9.${payload}.sig`;
```
ccp-bofai: `const header = Buffer.from(JSON.stringify({"alg":"HS256"})).toString("base64url");` and `return `${header}.${payload}.sig`; might be more readable.

packages/shared/src/auth/stores/__tests__/authStore.initialize.browser.test.ts
```diff
@@ -126,7 +126,10 @@ const FUTURE = Math.floor(Date.now() / 1000) + 3600;
 const PAST = Math.floor(Date.now() / 1000) - 60;
 
 function makeJwtPayload(claims: Record<string, unknown>): string {
-  const payload = Buffer.from(JSON.stringify(claims)).toString("base64url");
+  const payload = btoa(JSON.stringify(claims))
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=/g, "");
   return `eyJhbGciOiJIUzI1NiJ9.${payload}.sig`;
```
ccp-bofai: `eyJhbGciOiJIUzI1NiJ9` is not that readable, but I have seen this before in the tests... maybe you want to export this as HS256JwtHeader or something built from the actual text of the value? Mind you `sig` Does not look like it's valid for HS256. So maybe it's more "minimal JWT header". Given "sig" being the signature then maybe the header should specify alg as "none", I think that would be more correct is that an issue? I think the other instance of this uses Buffer's toString to get the output as URL safe Base64, that's probably an issue in a browser still but I believe that btoa is okay from Node 16 or so i.e. `node -e "console.log(btoa('{\"alg\":\"HS256\"}'))"`

packages/shared/src/auth/stores/__tests__/authStore.initialize.web.browser.test.ts
```diff
@@ -0,0 +1,264 @@
...
+const FUTURE = Math.floor(Date.now() / 1000) + 3600;
+const PAST = Math.floor(Date.now() / 1000) - 60;
+
+function makeJwtPayload(claims: Record<string, unknown>): string {
+  const payload = btoa(JSON.stringify(claims))
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=/g, "");
+  return `eyJhbGciOiJIUzI1NiJ9.${payload}.sig`;
+}
```
ccp-bofai: This probably should be a utility function usable from Node / Browser so using btoa should work.

ccp-raudur:
## Simplification review — line-count focus

Bug fixes look surgical and correct. Two big simplification opportunities below; the rest is minor.

### 🔴 1. Seven `authStore` test files duplicate ~14 `vi.mock()` blocks each

Lines of mock setup *before the first `describe`* per file:

| File | Mock-block lines |
|------|---:|
| `authStore.extensionLogin.test.ts` | 146 |
| `authStore.hydration.browser.test.ts` | 102 |
| `authStore.initialize.browser.test.ts` | 163 |
| `authStore.initialize.web.browser.test.ts` | 160 |
| `authStore.login.browser.test.ts` | 116 |
| `authStore.logout.extension.test.ts` | 121 |
| `authStore.tenantSwitch.test.ts` | 129 |
| **Total** | **937** |

All seven mock the same set: `#/auth/authConfig`, `#/auth/storageService`, `#/auth/userJwtSync`, `#/auth/userToJwtResponse`, `#/auth/utils/authStoreUtils`, `#/auth/getZkLoginAddress`, `#/auth/oauthTokenResponse`, `#/services/vaultService`, `#/stores`, `#/stores/tenantStore`, `#/utils`, `#/utils/tenantConfig`, `#/adapters`, `jose`. Only the mock-return shapes vary slightly.

**Fix:** Extract a shared `authStoreTestMocks.ts` — same pattern you already used for `keeperTestUtils.ts` (`vi.hoisted` handles + a `setupAuthStoreMocks()` helper). Conservative estimate: **−500 to −700 lines**, and every future authStore test gets much cheaper to write.

### 🟡 2. `apps/web/src/lib/routeUtils.ts` is unused — production routes still inline their own copies

The new file exports `getOptionalString`, `validateSearch`, `isRoutePath`, `resolveRoute` — but nothing in production imports it. The route files still define local copies:
- `apps/web/src/routes/index.tsx:7-22` — duplicates `resolveRoute`, `getOptionalString`, `validateSearch`
- `apps/web/src/features/auth/components/CallbackScreen.tsx:19` — duplicates `isRoutePath`

Result: duplicate code **plus** a 73-line test file that doesn't cover the live code path.

**Pick one:**
- Wire the two route files to import from `routeUtils.ts` (best — test then covers real code, removes ~20 lines of duplication)
- Or delete `routeUtils.ts` + `routeUtils.test.ts` entirely (−94 lines)

### 🟢 3. `useResponsive.ts` — `safeWidth` rename costs ~4 lines for one fallback

The guard at `useResponsive.ts:71-84` is correct, but renaming `width` → `safeWidth` ripples through three branches. Could fold to one reassignment at the top:

```ts
function getResponsiveState(width: number | undefined): ResponsiveState {
  width = typeof width === \"number\" && !Number.isNaN(width) ? width : 1200;
  // …rest unchanged
}
```

### 🟢 4. `authStore.ts` web-login try/catch wraps the non-zkLogin early-return unnecessarily

At `authStore.ts:289-326`, the `isZkLoginSuiChain(network)` early-return doesn't throw, so it doesn't need to be inside the `try`. Hoisting it above the `try` flattens one indentation level (~6 lines of body un-indent).

### What looks good
- The five bug fixes (`ROTATE_KEYPAIR` expiry extension in `keeper.ts` + `webVaultService.ts`, `runTenantSwitchCleanup` lock, `switchTenantAndReload` URL, `useResponsive` undefined-width, `deviceStore.localnetUrl` initial state) are all small, targeted, and well-justified.
- The keeper test rewrite (simulated → integration via `keeperTestUtils.ts`) is the right call even though net-positive in lines — it now tests real `keeper.ts`. `createKeeperTestContext()` + `setupKeeperSuite()` are exactly the pattern the authStore tests should adopt.
- Pre-commit → pre-push for `test:run` is the right move.
- Coverage regression check in CI is a nice safety net.

### Rough impact if (1)+(2) are applied
PR shrinks from **+4018** to roughly **+3300**, and the test code exercises more real production paths.

packages/shared/src/auth/stores/__tests__/authStore.tenantSwitch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks();
+    setupAuthStoreMocks(h);
+    useAuthStore.setState({ user: { id_token: "token" } as never });
+  });
+
+  it("runTenantSwitchCleanup clears JWTs, removes OIDC user, and clears zkLogin address cache", async () => {
+    await runTenantSwitchCleanup("stillness" as never);
+
+    expect(h.mockRemoveUser).toHaveBeenCalledTimes(1);
```
ccp-bofai: Please search the code base for calls to `toHaveBeenCalledOnce()` if there are any then please try to use  one or the other in the code base and not both. `tohaveBeenCalledOnce()` would be my preference. This is also something that could be done in a follow up PR.

packages/shared/src/auth/stores/__tests__/authStore.tenantSwitch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    await switchTenantAndReload("tauceti" as never);
+
+    expect(h.mockSetCurrentTenantId).toHaveBeenCalledWith("tauceti");
+    expect(reload).toHaveBeenCalledTimes(1);
+
+    Object.defineProperty(window, "location", {
+      configurable: true,
+      value: originalLocation,
+    });
```
ccp-bofai: This looks structurally wrong, mainly because it is setting state with no expectation after and seems to be doing what an "after" function would do, but it is not in a "finally" block so if an exception is thrown it is not guaranteed to be called. I would expect this block to be in a finally where the try starts immediately after the state variable that it is applying has been set.

packages/shared/src/auth/stores/__tests__/authStore.tenantSwitch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    // mockGetCurrentTenantId returns "stillness" and we pass "stillness" — early return
+    await switchTenantAndReload("stillness" as never);
+
+    expect(h.mockSetCurrentTenantId).not.toHaveBeenCalled();
+    expect(reload).not.toHaveBeenCalled();
+
+    Object.defineProperty(window, "location", {
+      configurable: true,
+      value: originalLocation,
+    });
```
ccp-bofai: Same issue as above. No following expectation and this is attempting to restore state but it is not guaranteed to be called.

packages/shared/src/stores/__tests__/deviceStore.rehydration.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  it("reconstructs the ephemeral public key from persisted bytes", async () => {
+    const publicKey = new Ed25519PublicKey(new Uint8Array(32).fill(7));
+
+    window.localStorage.setItem(
+      DEVICE_STORAGE_KEY,
+      JSON.stringify({
+        state: {
+          ephemeralPublicKeyBytes: Array.from(publicKey.toRawBytes()),
+          ephemeralPublicKeyFlag: KEY_FLAG_ED25519,
```
ccp-bofai: Possibly safer to use `publicKey.flag()` if that is returning the same value as this constant currently is. Main concern is that if the key type changes then this would also need changing. Alternatively define `expectedPublicKeyFlag` just after instantiating the `publicKey` and use that value instead of this constant.

packages/shared/src/stores/__tests__/deviceStore.rehydration.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    await useDeviceStore.persist.rehydrate();
+
+    const state = useDeviceStore.getState();
+    expect(state.ephemeralPublicKey?.toRawBytes()).toEqual(
+      publicKey.toRawBytes(),
+    );
+    expect(state.ephemeralPublicKeyBytes).toEqual(
+      Array.from(publicKey.toRawBytes()),
+    );
+    expect(state.ephemeralPublicKeyFlag).toBe(KEY_FLAG_ED25519);
```
ccp-bofai: See above

packages/shared/src/testing/testUtils.ts
```diff
@@ -0,0 +1,10 @@
+// ─── test utility functions ────────────────────────────────────────────────
+
+export function makeJwtPayload(claims: Record<string, unknown>): string {
+  const b64url = (obj: unknown) =>
+    btoa(JSON.stringify(obj))
+      .replace(/\+/g, "-")
+      .replace(/\//g, "_")
+      .replace(/=/g, "");
+  return `${b64url({ alg: "HS256" })}.${b64url(claims)}.sig`;
+}
```
ccp-bofai: Questionable function name. "payload" is what here? `makeJWT` I could understand.

packages/shared/src/utils/__tests__/buildTx.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        throw new Error("setSender failed");
+      }),
+      build: vi.fn(),
+    };
+
+    await expect(buildTx(tx as never, "0xabc", {} as never)).rejects.toThrow(
+      "setSender failed",
+    );
+    expect(tx.build).not.toHaveBeenCalled();
+  });
```
ccp-bofai: What exactly is this testing: https://github.com/MystenLabs/ts-sdks/blob/225f92819c08329058ce6ca5dcef83cc040d86fd/packages/sui/src/transactions/Transaction.ts#L227-L229 ? There are various other tests in this PR that are somewhat questionable as to what exactly they are testing.
I do think it would be better to have a test that illustrates what happens when the Sender is already set in the Transaction to something other than the current wallet address.

packages/shared/src/wallet/__tests__/zkProof.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      proofPoints: {
+        a: ["1", "2"],
+        b: [
+          ["3", "4"],
+          ["5", "6"],
+        ],
+        c: ["7", "8"],
+      },
+    },
+  };
```
ccp-bofai: Is this supposed to be the same as the response Enoki's `zkp` API? i.e. https://docs.enoki.mystenlabs.com/http-api/openapi#create-zklogin-zkp or the example at the end of this section: https://docs.sui.io/sui-stack/zklogin-integration/#get-the-zero-knowledge-proof A few properties seem to be missing.

packages/shared/src/wallet/__tests__/zkProof.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    const fetchMock = vi
+      .fn()
+      .mockResolvedValueOnce({
+        ok: true,
+        json: vi.fn().mockResolvedValue(testnetProof),
+      })
+      .mockResolvedValueOnce({
+        ok: true,
+        json: vi.fn().mockResolvedValue(mainnetProof),
+      });
```
ccp-bofai: The title suggests something about routing being tested. What exactly is this testing? Looks like it is testing that mockResolvedValueOnce works rather than testing routing. I would expect this to break if the calls to fetchZkProof are reordered or the calls to mockResolvedValueOnce are reordered.

packages/shared/src/wallet/__tests__/zkProof.test.ts
```diff
@@ -0,0 +1,212 @@
...
+  });
+
+  it("throws when Enoki returns a non-OK response with a non-JSON body", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn().mockResolvedValue({
+        ok: false,
+        status: 500,
+        statusText: "Server Error",
+        json: vi.fn().mockRejectedValue(new Error("not json")),
```
ccp-bofai: This should be a SyntaxError as [fetch returns a Response](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). [`Response.json` instance method documents the return for the circumstance that this test describes](https://developer.mozilla.org/en-US/docs/Web/API/Response/json)

packages/shared/src/wallet/__tests__/zkProof.test.ts
```diff
@@ -0,0 +1,212 @@
...
+        idToken: "id-token",
+        enokiApiKey: "enoki-api-key",
+      }),
+    ).rejects.toThrow("Failed to fetch ZK proof");
+  });
+
+  it("re-throws when fetch itself rejects with a network error", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn().mockRejectedValue(new Error("Network failure")),
```
ccp-bofai, line 222: Why not use `Response.error()` here (it would be more correct and perhaps the fetchZkProof is expecting something of that shape)? It would match up more with the `fetch` API. https://developer.mozilla.org/en-US/docs/Web/API/Response/type#error seems to be a good place for details here.

bunfig.toml
ccp-bofai, line 1: GitHub is showing the "no entry" sign at the end of this file ;)

ccp-bofai (changes requested):
There's still some odd looking things going on in some of the tests. There's also some comments that should be dealt with in later PRs too, so I am not expecting changes in this PR for all of them.

packages/shared/src/wallet/__tests__/zkProof.test.ts
fcsondheim: updated to add missing properties and comply to the expected type.

packages/shared/src/wallet/__tests__/zkProof.test.ts
fcsondheim: `Response.error()` would be expected as a return from the API server. In the case of a network error there would likely not be a server `Response` at all, whether `ok` or `error`.

apps/web/src/lib/__tests__/tenantUrlSync.test.tsx
```diff
@@ -0,0 +1,220 @@
...
+      tenantId: DEFAULT_TENANT,
+      changed: false,
+    });
+  });
+
+  afterEach(() => {
+    Object.defineProperty(window, "location", {
+      value: originalLocation,
+      configurable: true,
+    });
```
ccp-bofai: I am seeing this block of code repeated quite a bit in this file. Might be an idea to a function that does this called `unsetMockLocation` (seeing that it is undoing `setMockLocation`)

packages/shared/src/auth/stores/__tests__/authStore.logout.test.ts
```diff
@@ -134,15 +134,15 @@ describe("authStore.logout()", () => {
 
     await useAuthStore.getState().logout();
 
-    expect(mockClear).toHaveBeenCalledTimes(1);
+    expect(mockClear).toHaveBeenCalledOnce;
```
ccp-bofai: I believe this is probably not the intention. See: https://vitest.dev/api/expect.html#calledonce where it points out that the Jest style (as above) is a function and should have parenthesis. If this is true you probably want to check for this in other places too.

packages/shared/src/auth/stores/__tests__/authStore.tenantSwitch.test.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+  runTenantSwitchCleanup,
+  switchTenantAndReload,
+  useAuthStore,
+} from "#/auth/stores/authStore";
+import { setWindowLocation } from "#/testing";
+
+describe("tenant switch auth cleanup", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
```
ccp-bofai: Seeing this in `beforeEach` makes me think that we have tests where mocks are not being cleared after the test and that side effects are expected. In an ideal world this should not be needed and there should be an `afterEach` doing this. Might be worth have a look if we are doing this in every describe's beforeEach rather than in some beforeEach and some afterEach as mixing them might introduce side effects. I think the main thing here is a consistent pattern that is not going to introduce side effects when only certain tests are run and when all tests are run.

packages/shared/src/auth/stores/__tests__/authStore.tenantSwitch.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  it("runTenantSwitchCleanup does not lock the vault", async () => {
+    await runTenantSwitchCleanup("stillness" as never);
+
+    expect(h.mockDeviceLock).not.toHaveBeenCalled();
+  });
+
+  it("switchTenantAndReload updates currentTenantId then reloads the page", async () => {
+    const reload = vi.fn();
+    const originalLocation = window.location;
```
ccp-bofai: I would probably put this at the start of the current describe block so that 92 and the repeated line at 107 is not required.

packages/shared/src/wallet/__tests__/zkProof.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    warn: vi.fn(),
+    error: vi.fn(),
+  }),
+}));
+
+import { fetchZkProof } from "#/wallet/zkProof";
+
+const headerBase64 = Buffer.from(
+  JSON.stringify({ alg: "none", typ: "JWT" }),
+).toString("base64url");
```
ccp-bofai: Not in this PR but something to consider... This is similar to `makeJWT` but it looks to be a bit NodeJS only, unlike `makeJWT`. You could possibly use `makeJWT` and split on the `.` in the output and use the first segment.

packages/shared/src/wallet/__tests__/zkProof.test.ts
```diff
@@ -0,0 +1,236 @@
...
+}));
+
+vi.mock("#/utils/logger", () => ({
+  createLogger: () => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  }),
+}));
```
ccp-bofai, line 20: Not in this PR... I seem to be remembering this mock from other files and that it is essentially repeated boiler plate code. Might be an idea to look over the mocks at the start of the files and identify common ones and have a fairly standard set of mocks.

packages/shared/src/wallet/__tests__/zkProof.test.ts
ccp-bofai: Good point. Looks like its actually TypeError that's returned if you request a valid looking URL that is not listening.. `node --eval "try { await fetch('http://[::1]:666/does-not-exist') } catch (error) { console.log(error) }"` So `Error` is fine as its only checking its `message` property.

ccp-bofai (approved):
Please read the comments for suggested follow up 
Looks good to me

PR #106, opened 2026-05-19
chore: refactor to reduce complexity

fcsondheim:
Refactors the keeper and auth store code to reduce large, high-complexity files

Breaks up two large modules into smaller responsibility-focused files:
- `keeper.ts` is reduced to a thin offscreen-document entrypoint that registers the Chrome message listener and sends the ready message.
- Keeper logic is split into dedicated modules for state, crypto/session handling, vault operations, zkProof handling, localnet handling, and message dispatch.
- `authStore.ts` is reduced to Zustand store wiring, hydration, tenant switching, and public store exports.
- Auth workflows are split into smaller modules for extension initialization, extension login, Chrome auth handling, web login/init, logout cleanup, user session conversion, and shared workflow utilities.
- Expand testing for authStore behaviour


apps/extension/entrypoints/keeper/keeperState.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export function unlockVaultWithKeypair(keypair: Ed25519Keypair): void {
+  ephemeralKey = keypair;
+  _vaultUnlocked = true;
+  _vaultUnlockExpiry = Date.now() + VAULT_UNLOCK_MS;
+}
+
+export function replaceEphemeralKey(keypair: Ed25519Keypair): void {
+  ephemeralKey = keypair;
+  _vaultUnlocked = true;
```
ccp-bofai: I might be wrong but I think this line should not be here. The similarity to the above function suggests that it is extra to me.

apps/extension/entrypoints/keeper/keeperState.ts
```diff
@@ -0,0 +1,107 @@
...
+    return true; // Already locked
+  }
+
+  if (_vaultUnlockExpiry && Date.now() > _vaultUnlockExpiry) {
+    lockVault();
+    return true; // Now locked
+  }
+
+  return false; // Still unlocked
+}
```
ccp-bofai: Seeing the name of the function start with "check" I would expect it to throw if its expectations are not met. However, it returns a boolean so it looks more like it could be an "is", "has" or "should" prefix but I don't think that is the case. I am thinking that this would be clearer if it was simply named `enforceExpiry`.

apps/extension/entrypoints/keeper/keeperState.ts
fcsondheim: Will remove the vaultUnlocked line; however, the key replacement line itself is distinct from the unlock function above (the previous function also triggers the unlock expiry timer to update).

packages/shared/src/auth/stores/__tests__/authUserSession.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  buildUserFromOAuthResponse,
+  jwtTiming,
+  persistEnrichedUser,
+} from "#/auth/stores/authUserSession";
+import { makeJwt } from "#/testing";
+import type { JwtResponse, OAuthTokenResponse } from "#/types/authTypes";
+
+function makeJwtResponse(overrides: Partial<JwtResponse> = {}): JwtResponse {
+  return {
+    id_token: makeJwt({ sub: "user-1", iat: 1000, exp: 4600 }),
```
ccp-bofai: You might want to move this `{ sub: "user-1", iat: 1000, exp: 4600 }` to a constant outside of this function so that it can be used at line 59 also.

ccp-bofai (approved):
Some comments but looks good to me.

PR #109, opened 2026-05-21
chore: semicolons as needed

fcsondheim:


ccp-bofai (approved):
Looks good to me. Does look like some ASI happened and that biome is playing it safe in regards to code where some weird ambiguous interpretation could occur, which is okay I think, possibly even good.

PR #111, opened 2026-05-28
chore: use gh packages version of `@evefrontier/dapp-kit`

fcsondheim:
use github packages version of `@evefrontier/dapp-kit`

Resolves the previous split between packages from the same org on different registries; `dapp-kit` was being resolved from npmjs and `wallet-core` was being resolved from github registry. This PR reads both same-org packages from the same github registry.


ccp-raudur (approved):
LGTM

PR #113, opened 2026-06-01
feat: migrate OAuth to PKCE, remove client secrets

fcsondheim:
Migrates the extension and web auth flows from confidential-client (shared secret) to public-client PKCE (RFC 7636), removing client secrets from the codebase entirely.

**PKCE**
- `oauthService.ts`: `getAuthRequest()` generates a `(codeVerifier, codeChallenge)` pair per auth request via Web Crypto API; replaces `getAuthUrl()`
- `exchangeCode.ts`: sends `code_verifier` instead of `client_secret`
- `authConfig.ts`: `UserManager` drops `client_secret`; `oidc-client-ts` handles PKCE on the web flow automatically
- `VITE_TENANT_*_CLIENT_SECRET` env vars removed from `.envsample`, type declarations, `TenantConfig`, and all tenant entries in `constants.ts`

**Dependency**
- Replaces `@evefrontier/dapp-kit` with `@evefrontier/wallet-core@0.0.1-alpha.6` for `TenantId` and tenant utilities

**Bug fixes in `dappLogin.ts`**
- `launchWebAuthFlow` callback was synchronous with a floating `.then().catch()` — converted to `async`/`try-catch`
- `getCurrentTenantId()` was re-read inside the callback; now closes over the tenant captured before the flow starts, preventing verifier/endpoint mismatch if the user switches tenants mid-popup
- Token-exchange `.catch` was sending a message without an `id` field, causing the dapp tab to hang on failure

**CI security**
- Pin `oven-sh/setup-bun` and `davelosert/vitest-coverage-report-action` to commit SHAs
- Fix template injection in `release-extension.yml`: `github.event.release.tag_name` passed via env var instead of inline shell expansion


apps/extension/src/lib/background/services/oauthService.ts
```diff
@@ -1,7 +1,35 @@
...
+import { sha256 } from '@evevault/shared/utils'
 
-function getAuthUrl(params: { tenantId: TenantId; nonce: string }) {
+function base64UrlEncode(bytes: ArrayBuffer | Uint8Array): string {
+  const input = bytes instanceof Uint8Array ? bytes : new Uint8Array(bytes)
+  return btoa(String.fromCharCode(...input))
+    .replaceAll('+', '-')
+    .replaceAll('/', '_')
+    .replaceAll('=', '')
+}
```
ccp-bofai: This looks like a fairly general utility function and feels a bit odd in this file rather than a file that deals with base64 URL encoding / decoding?

fcsondheim: Good point! Yes I'll move that to a better place

ccp-bofai (approved):
LGTM

PR #115, opened 2026-06-03
feat: address qlty issues - shared

fcsondheim:
Refactors the shared package to reduce complexity by extracting large hooks, store actions, and utility logic into focused helper modules.

- Split device/context store initialization, chain switching, lock, proof, and rehydration logic into dedicated helpers.
- Broke down wallet hooks for balance loading, token sending, gas estimation, validation, refetching, transaction construction, and signing context callbacks.
- Moved logger internals into `logger.helpers.ts` while keeping `createLogger` as the public entry point.
- Simplified zk signing, transaction parsing, coin metadata, `useDevice`, `keeperService`, and `parseTxBytes` by separating reusable logic from orchestration code.


packages/shared/src/stores/deviceStore/actions/initChainHelpers.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  setChainData(set, get, chain, {
+    maxEpoch: numericMaxEpoch.toString(),
+    maxEpochTimestampMs,
+    nonce,
+    jwtRandomness,
+  })
+}
+
+const NULL_EPOCH = { maxEpoch: null, maxEpochTimestampMs: null } as const
```
ccp-bofai: This seems a bit weirdly placed in this file. Its usage is further up. Would have expected something like this near the top of the file possibly before `log` is defined.

packages/shared/src/stores/deviceStore/actions/initDeviceHelpers.ts
```diff
@@ -0,0 +1,209 @@
...
+import type { GetDeviceState, SetDeviceState } from './types'
+
+const log = createLogger()
+
+type InitActionParams = {
+  pin: string
+  currentChain: SuiChain
+  set: SetDeviceState
+  get: GetDeviceState
+}
```
ccp-bofai: Slightly odd ordering, isn't it? Would have expected to see `log` be defined after the types are defined at file scope.

packages/shared/src/stores/deviceStore/actions/initStateHelpers.ts
```diff
@@ -0,0 +1,101 @@
...
+  return Boolean(
+    isPresent(data?.nonce) &&
+      isPresent(data?.maxEpoch) &&
+      !isDeviceDataExpired(data),
+  )
+}
+
+const isPresent = <T>(value: T | null | undefined): value is T => {
+  return value !== null && value !== undefined
+}
```
ccp-bofai: Could this be defined far higher up before it's used. It looks quite generic too, i.e. there might be other functions with the same signature and functional behaviour but with a different name in the res of the code base.

packages/shared/src/stores/deviceStore/rehydrationHelpers.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const publicKey = reconstructPublicKey(
+    state.ephemeralPublicKeyBytes,
+    state.ephemeralPublicKeyFlag ?? null,
+  )
+
+  if (publicKey) {
+    state.ephemeralPublicKey = publicKey
+    log.debug(
+      `Reconstructed ${isWeb() ? 'Secp256r1' : 'Ed25519'} public key from storage`,
```
ccp-bofai: I believe you can use `flag()` on a Mysten PublicKey to get a number that represents the type of signature scheme i.e. Ed25519, Secp256r1, Secp256k1 and so on. I am not sure if the type of `publicKey` is one of those, so this might not be relevant.

packages/shared/src/stores/contextStore.switching.ts
```diff
@@ -0,0 +1,182 @@
...
+    await initializeDeviceDataIfNeeded(chain)
+
+    set({ loading: false })
+    log.info('Successfully switched to zkLogin chain', { chain })
+    return SWITCH_SUCCESS
+  } catch (error) {
+    log.error('Failed to complete network switch', error)
+    set({ loading: false })
+    set({ chain: previousChain })
+    return { success: false, requiresReauth: false }
```
ccp-bofai: The returns here feel a bit unclear. I guess that the value that `SWITCH_SUCCESS` has was repeated in this file and thus turned into a constant and that the other returns with the same shape are not repeated and have not been turned into constants (maybe they could but it's questionable defining them if they are only used once). However, at the same time `SWITCH_SUCCESS` is describing the post state and I see line 112 that logs the text "successfully switched...", which makes me wonder if `SWITCH_SUCCESS` should be `SWITCHED_SUCCESSFULLY` but that could just be `SWITCHED` but maybe `SWITCHED_REAUTH_NOT_REQUIRED` might be better as it is more suggestive of what is being returned rather than leaving the state of `requiresReauth` as a bit of an unknown.

packages/shared/src/stores/contextStore.ts
```diff
@@ -3,19 +3,21 @@ import { SUI_TESTNET_CHAIN, type SuiChain } from '@mysten/wallet-standard'
...
+  checkNetworkSwitchRequirement,
+  forceSetContextChain,
+  setContextChain,
+} from './contextStore.switching'
 
 // Store dependency direction is intentional: contextStore may coordinate with
 // deviceStore during network switches, but deviceStore must not import contextStore.
-const log = createLogger()
-
 const INITIAL_TENANT_ID = 'stillness' as TenantId
```
ccp-bofai: At some point (in the future) we might want to have the `'stillness'` value used here imported from some sort of well named constants set for the whole code base. This would probably make it easier for anyone way off in the future that wants to do their own thing with this code and they want to do it with something after 'stillness'. Having hardcoded values like this might make them curse at us :) I could imagine something along the lines of `import { CONTEXT_STORE_INITIAL_TENANT_ID } from 'systemDefaults'`. Also if someone wanted to fork this for their own project this might be an issue.

packages/shared/src/wallet/hooks/useBalance.helpers.ts
```diff
@@ -0,0 +1,225 @@
...
+
+const resolveBalanceMetadata = (
+  meta: BalanceAndMetadataResponse['coinMetadata'] | undefined,
+  coinType: string,
+): BalanceMetadata | null => {
+  if (coinType === SUI_COIN_TYPE) {
+    return DEFAULT_SUI_METADATA
+  }
+  if (isEveCoinType(coinType)) {
+    return DEFAULT_EVE_TESTNET_METADATA
```
ccp-bofai, line 199: This might be correct today but.... it might be better if this sort of resolution was more centralised.

fcsondheim: This will be done in a separate PR as this would be a behaviour change rather than a refactor :)

packages/shared/src/wallet/hooks/useTransactionHistory.ts
```diff
@@ -97,9 +97,7 @@ export function useTransactionHistory({
       )
 
       const transactions = parsed
-        .filter(
-          (tx): tx is import('#/types/components').Transaction => tx !== null,
-        )
+        .filter(isPresent)
```
ccp-bofai: I do wonder if it is obvious that "isPresent" means "is not undefined etc" or something similar to that? I think isPresent should probably have a name that describes what it actually checks even if it is something like isNotUndefinedAndIsNotNull rather than the slightly fluffy "present", but if there is somewhere, where that name would make sense then "import as" for that accordingly. I think the jump to "it's a transaction" is simpler from "it's not undefined and not null" rather than "is present". Some people might wonder if we have a Present class defined at the bottom of the file. Perhaps this should go into a separate PR for better review/clarity because I think I have seen a few of these?

packages/shared/src/wallet/utils/parseTransaction.ts
```diff
@@ -49,136 +49,213 @@ export async function parseGraphQLTransaction(
...
+}
 
-  const metadata = await fetchCoinMetadata(graphqlClient, coinType)
-  const decimals = metadata?.decimals ?? 9
+const isNonSuiAmountChange = (change: GraphQLBalanceChange): boolean => {
+  return getCoinType(change) !== SUI_COIN_TYPE && hasAmount(change)
+}
+
+const hasAmount = (change: GraphQLBalanceChange): boolean => {
+  return isPresent(change.amount)
```
ccp-bofai: More "isPresent" ambiguity

ccp-bofai (approved):
Please go over the comments and pull out what you think are actually things we should deal with in future Jira/PRs

PR #116, opened 2026-06-03
feat: address qlty issues - components

fcsondheim:
Refactors shared UI components to reduce duplication and keep rendering, state hooks, and helper logic focused.

- Split larger components into dedicated helper and parts modules for dropdowns, network selection, token lists, toasts, and the transactions screen.
- Extracted `TokenRow` and token list refresh/copy/remove logic out of `TokenSection`.
- Added a shared `createSvgIcon` factory and migrated repeated icon SVG boilerplate to use it.
- Reworked dropdown identicons to render from compact pattern data instead of duplicated SVG markup.
- Preserves the existing component behavior while making the shared component layer easier to maintain.


ccp-bofai (approved):
LGTM

PR #118, opened 2026-06-06
chore: move functions to import from wallet-core

fcsondheim:
Removes implementations from `@evevault/shared` and replaces
them with canonical imports from `@evefrontier/wallet-core`.

## What moved

| Local | Now imported from |
|---|---|
| `utils/buildTx.ts` → `buildTx` | `wallet-core/utils` → `buildTransactionBytes` |
| `wallet/eveToken.ts` → `getEveCoinType`, `isEveCoinType` | `wallet-core/definitions` |
| `wallet/signWithIntent.ts` | `wallet-core/crypto` |
| `wallet/zkSignAny.helpers.ts` → `loadZkProof`, `createZkLoginSignature` | `wallet-core/crypto` |

## Behavioural changes

- **`signWithIntent`** — no longer wraps keypair errors as `'Error signing message'`;
  raw errors now propagate.
- **`loadZkProof`** — wallet-core's version returns `PartialZKLoginSignature`
  directly (unwrapped from the response envelope), so the `zkSignAny` call site
  drops `.data`.
- **`isEveCoinType`** — wallet-core does not include the pre-per-tenant-package
  legacy coin type
  (`0x59d7bb2e0feffb90cb2446fb97c2ce7d4bd24d2fb98939d6cb6c3940110a0de0::EVE::EVE`).
  **This is safe: legacy EVE is fully out of circulation, so no live balances depend on it.** The removal is now asserted in `eveToken.test.ts` so it reads as deliberate and any accidental re-introduction is caught.

## Cleanup

- Dropped the now-dead `sui_address` field from the `ZK_EPH_SIGN_BYTES` message path. The keeper's `handleEphSign` no longer reads it (it signs via the proofed ephemeral keypair), mirroring the localnet sign path which already dropped its address.


apps/extension/entrypoints/keeper/__tests__/keeper.operations.test.ts
```diff
@@ -513,7 +574,57 @@ describe('Keeper LOCALNET_SET_KEYPAIR handler', () => {
 
     expect(resp.ok).toBe(false)
     expect(resp.error).toBe('No localnet keypair loaded')
-    expect(mockSignWithIntent).not.toHaveBeenCalled()
+    expect(mockSignTransaction).not.toHaveBeenCalled()
+  })
+
+  it('signs transaction bytes with TransactionData scope', async () => {
+    const keypair = Ed25519Keypair.generate()
```
ccp-bofai, line 581: Should this be ZKEd25519Keypair?

apps/extension/entrypoints/keeper/__tests__/keeper.operations.test.ts
```diff
@@ -513,7 +574,57 @@ describe('Keeper LOCALNET_SET_KEYPAIR handler', () => {
...
+    })
+
+    expect(resp.ok).toBe(true)
+    expect(resp.bytes).toBe('txBytes')
+    expect(resp.signature).toBe('txSig')
+    expect(mockSignTransaction).toHaveBeenCalledWith(new Uint8Array([1, 2, 3]))
+  })
+
+  it('signs personal message bytes with PersonalMessage scope', async () => {
+    const keypair = Ed25519Keypair.generate()
```
ccp-bofai, line 605: Should this be ZKEd25519Keypair?

apps/extension/entrypoints/keeper/__tests__/keeper.operations.test.ts
```diff
@@ -605,7 +716,7 @@ describe('Keeper UNLOCK_VAULT — localnet key restoration', () => {
   it('leaves localnetKey null when wrong PIN is used to decrypt', async () => {
     const localnetKeypair = Ed25519Keypair.generate()
```
ccp-bofai, line 717: Should this be ZKEd25519Keypair?

apps/extension/entrypoints/keeper/keeperCrypto.ts
```diff
@@ -1,3 +1,4 @@
+import { ZKEd25519Keypair } from '@evefrontier/wallet-core/crypto'
 import { decrypt, deriveAesKey, type HashedData } from '@evevault/shared'
 import { Ed25519Keypair } from '@mysten/sui/keypairs/ed25519'
```
ccp-bofai, line 3: Why is this import here? Seems a bit odd importing ZKEd25519Keypair and Ed25519Keypair

packages/shared/src/utils/__tests__/buildTx.test.ts
ccp-bofai, line 1: Is this test file needed or have the tests not been moved to wallet-core?

packages/shared/src/wallet/__tests__/eveToken.test.ts
ccp-bofai, line 1: Has this test file not been moved over to wallet-core?

packages/shared/src/wallet/zkSignAny.helpers.ts
```diff
@@ -86,77 +68,45 @@ export const requireZkLoginClaims = (
...
-  const result = await signWithIntent(msgBytes, scope, {
-    sui_address: user.profile?.sui_address as string,
-    keypair: signer,
-  })
-  return { bytes: result.bytes, userSignature: result.userSignature }
+  if (!user.profile?.sui_address) {
+    throw new Error('[signWithWebEphemeralKey] User address not found')
+  }
+
+  signer.applyZKProof(zkProofData)
```
ccp-bofai, line 86: I am a little surprised to `applyZKProof` here rather than being called when the data has become available at the end of the ZK Login when the proof data is available and the key would exist because its public part is used in the process (so the proof data could be applied to the ephemeral key at that point in time, and then you would not have to be adding `zkProofData: ZKProofData` in various signing methods).

apps/extension/entrypoints/keeper/vaultHandlers.ts
```diff
@@ -148,20 +147,21 @@ export function handleEphSign(
...
-        {
-          sui_address: sui_address as string,
-          keypair: key,
-        },
-      )
+      const { msgBytes, scope, zkProofData } = message
+      if (!zkProofData) {
+        throw new Error('[KEEPER_EPH_SIGN] zkProofData is required')
+      }
+      key.applyZKProof(zkProofData as ZKProofData)
```
ccp-bofai, line 154: This feels a bit odd here (same as `packages/shared/src/wallet/zkSignAny.helpers.ts`) where the proof data is applied just before signing rather than when it becomes available as part of the ZK Login process.

ccp-bofai (commented):
This is looking good :) Got a few comments around proof data handling/applying. Also noticed some test files that are only testing things that are in wallet-core (the tests probably should be in wallet-core not here, maybe they are? so maybe those test files just need removing?)

apps/extension/entrypoints/keeper/__tests__/keeper.operations.test.ts
fcsondheim: Not this one; this is being called for localnet which is non-zk

apps/extension/entrypoints/keeper/__tests__/keeper.operations.test.ts
fcsondheim: Not this one; this is being called for localnet which is non-zk

apps/extension/entrypoints/keeper/__tests__/keeper.operations.test.ts
fcsondheim: Not this one; this is being called for localnet which is non-zk

apps/extension/entrypoints/keeper/keeperCrypto.ts
fcsondheim: this uses ZKEd25519Keypair for the networks that support zk, and Ed25519Keypair for the localnet path

packages/shared/src/wallet/zkSignAny.helpers.ts
fcsondheim: applying the data at the end of zklogin is premature in a multi-network environment. `zkproof` differs across the different networks (mainnet, devnet, testnet) so switching between them requires different zkproofs to be applied depending on which chain is being signed for

apps/extension/entrypoints/keeper/vaultHandlers.ts
fcsondheim: applying the data at the end of zklogin is premature in a multi-network environment. `zkproof` differs across the different networks (mainnet, devnet, testnet) so switching between them requires different zkproofs to be applied depending on which chain is being signed for

ccp-bofai (approved):
LGTM

PR #119, opened 2026-06-09
test: expand test coverage

fcsondheim:
Cleans up the zkLogin ephemeral-key signing path, consolidates the base64 codec used by the key-encryption layer, removes dead code, and adds substantial test coverage across shared and app code.

## Signing flow
- **Drop the unused `sui_address` field** from the `ZK_EPH_SIGN_BYTES` message flow. It was passed from `signWithExtensionEphemeralKey` into the keeper but never read. Removing it trims the message payload end to end.
- **Document the inline scope routing.** Added comments in both `handleEphSign` (keeper) and `signWithWebEphemeralKey` (shared) explaining why we call `signTransaction`/`signPersonalMessage` directly after `applyZKProof` rather than routing through wallet-core's `signWithIntent`, and why `applyZKProof` is safe to re-apply on every sign (idempotent; picks up a fresh proof on mid-session epoch rollover).

## Base64 codec
- **Extracted a single-source base64 codec** (`utils/keys/base64.ts`) with `bytesToB64`/`b64ToBytes`, now shared by `encrypt`, `decrypt`, and key derivation. Using one `atob`/`btoa`-based implementation everywhere prevents mismatched codecs from silently corrupting stored ciphertext.

## Dead code removal
- Removed code whose only references were its own tests:
  - `formatTransactionAmount` (`wallet/utils/formatTransaction.ts`) — kept the still-used `extractSymbolFromCoinType`.
  - The entire `RouteContext` machinery (`apps/web/.../routeContext.tsx`) — `useRouteContext` was the only consumer, so the provider and its `__root.tsx` wrapper went too.
  - `needsPersistedRehydration` (`deviceStore/actions/initStateHelpers.ts`) and its barrel re-export.

## Test coverage
- Adds new suites for `getZkLoginAddress`, `zkJwt`, `networks`, `calculate`, `devMode`, `logger` helpers, `useBalance` helpers, `useSendToken` gas/validation, `balanceMetadata`, `coinTypeFormat`, `parseTransaction`, and the new key/decrypt/derive paths.
- Expands `authConfig` and `authStoreUtils` tests; relocates `vaultService`/`webVaultService` tests into `__tests__` directories.
- Updates vault-handler tests to drop the removed `sui_address` field.


packages/shared/src/utils/format.ts
```diff
@@ -60,7 +60,7 @@ export function toSmallestUnit(amount: string, decimals: number): bigint {
 }
 
 /** SUI decimals (1 SUI = 10^9 MIST). */
-const SUI_DECIMALS = 9
+export const SUI_DECIMALS = 9
```
ccp-bofai: is this not the same as `SUI_DECIMALS` from `@mysten/sui/utils` ? Wondering if it should just be a re-export of that? https://sdk.mystenlabs.com/typedoc/variables/_mysten_sui.utils.SUI_DECIMALS.html

packages/shared/src/utils/__tests__/format.node.test.ts
ccp-bofai, line 1: These tests (or at least the ones being show in the GitHub UI in this PR) are all using SUI_DECIMALS. This makes me wonder if this is a parameter to the functions being unit tested that is effectively constant, so it being a parameter looks questionable. Maybe it's different for different coins? It does not seem to get tested with other values so I am wondering if either: (1) there are tests testing other values than SUI_DECIMALS, in which case this is fine. (2) that SUI_DECIMALS is the only value ever to be passed and the parameter could be baked into the function with a slight rename to indicate that it is SUI specific.

packages/shared/src/utils/base64.ts
ccp-bofai, line 1: I see this is under a directory called `keys` . I don't see this code as being something specific to "keys" so I would expect it to be perhaps a directory higher or in a more obvious place e.g. "codecs" perhaps?

ccp-bofai (commented):
Some comments, would be good to discuss I think

packages/shared/src/utils/__tests__/format.node.test.ts
fcsondheim: Most coins will default to SUI_DECIMALS, but it's not a rule that each coin's smallest unit has to be to the 9th decimal. But it's a good idea to add a test for values other than SUI_DECIMALS.

ccp-bofai (approved):
LGTM

PR #121, opened 2026-06-10
fix(extension): harden dapp auth message bridge boundary

fcsondheim:
Tightens the extension's messaging boundary so page↔background traffic is authenticated, schema-validated, and origin-scoped, and so wallet responses to pages carry only the minimum metadata.

### Inbound (page → background)
- **Origin check** — page messages are accepted only when `event.source === window` and `event.origin === window.location.origin`.
- **Allow-list + schema validation** — only `connect` and the public Wallet-Standard sign actions are forwarded, each shape-checked; anything else is dropped at the content script.
- **Sender authorization** — the background tags each route as `dapp` or `extension` and rejects messages whose sender isn't authorized for that route.
- **Single connect entry point** — consolidated to one `connect` message (removed a duplicate path).

### Outbound (background → page)
- Responses are posted to the specific page origin (never a wildcard target).
- A single `sendToTab()` choke point is the only path to the page; its `TabBoundMessage` type plus a runtime check ensure responses contain only wallet metadata (`{ chain, address, publicKey }`) and no token material.
- Outbound objects are built by explicit allow-list rather than spreading upstream responses.

### zkLogin resolution
- Account lookup moved into the background; the page receives only the resolved `{ address, publicKey }`.


ccp-bofai (approved):
LGTM

PR #122, opened 2026-06-10
fix(extension): per-origin dApp permissions, signing request context, and log hardening

fcsondheim:
Hardens the extension's dApp signing flow so users can see and authorise what they're approving, and tightens logging:
- sites must connect before they can request signatures, scoped per origin and per chain;
- every signing prompt shows the verified requesting site, account, and request details;
- disconnecting clears a site's permission;
- sensitive values are kept out of logs;
- a sponsored-transaction field is typed consistently.

### Per-origin dApp connection permissions 
- Connecting a site records a permission keyed by its **origin** (derived from the trusted message sender, not from page-supplied data) and the chain it connected on.
- `signTransaction` / `signPersonalMessage` / `signAndExecuteTransaction` / sponsored-transaction requests are rejected unless the requesting origin is connected on the active chain, with a clear "connect first" error.
- Disconnecting a site (and resetting the vault) clears its stored permission.
- The permission/connect flow was consolidated to reduce duplication.

### Signing request context in prompts
- Every signing prompt now displays the **verified requesting site**, the **account**, the **request type**, and when the site connected — so the user can confirm the source before approving.
- Sponsored-transaction prompts additionally show the action, assembly, type, name, URL, and description.
- The requesting context is persisted on the pending action and read by the prompt.

### Log hardening
- PINs, encrypted key material, and zkProof objects are no longer passed to the logger.
- Added `redactSensitive()` (+ `SENSITIVE_FIELDS`) in `@evevault/shared/utils`, used to sanitize arbitrary/unknown message payloads before logging, plus a regression test that fails if a sensitive value reaches a logger call.

### Type fix
- `PendingSponsoredTransaction.assembly` is now `number` (the `assemblyId`), matching the message type; the value is stringified only at the display row.

Screenshot 2026-06-15 at 14 51 32


apps/extension/src/lib/background/services/dappPermissions.ts
```diff
@@ -0,0 +1,242 @@
...
+    ...(stringOrUndefined(sender.tab?.title) && { title: sender.tab?.title }),
+    ...(stringOrUndefined(sender.tab?.favIconUrl) && {
+      favIconUrl: sender.tab?.favIconUrl,
+    }),
+  }
+}
+
+function isSuiChainArray(value: unknown): value is SuiChain[] {
+  return Array.isArray(value) && value.every((item) => typeof item === 'string')
+}
```
ccp-bofai, line 79: I am wondering if this could be a bit tighter (`sui:*` and "localnet" for example) because this looks like `['', '', '']` would be okay

ccp-bofai (approved):
Looks good to me

PR #127, opened 2026-06-15
security fixes

fcsondheim:
Consolidates four security hardening PRs (#121, #122, #123, #128) into a single reviewable diff, plus a SECURITY.md and a manifest fix.

### 1. Harden the dApp ↔ background message bridge (#121)

The content-script/background IPC accepted messages from any source and forwarded them without schema checks.

- **Origin check** — page messages accepted only when `event.source === window` and `event.origin === window.location.origin`; cross-origin drops are tested.
- **Allow-list + schema validation** — only `connect` and the public Wallet-Standard sign actions are forwarded; anything else is dropped at the content script.
- **Sender authorization** — the background tags each route as `dapp` or `extension` and rejects messages whose sender isn't authorized for that route.
- **Outbound scoping** — responses are posted to the specific page origin (never `*`); a `sendToTab()` choke point ensures responses carry only wallet metadata (`{ chain, address, publicKey }`) and no token material.
- **Single connect entry point** — duplicate connect path removed.

### 2. Per-origin dApp connection permissions + signing request context + log hardening (#122)

Sites could call `signTransaction` without ever connecting, and signing prompts showed no information about who was asking.

- **Per-origin permissions** — connecting records a permission keyed by origin (from the trusted message sender, not page-supplied data) and chain. Sign requests are rejected with a "connect first" error if the origin isn't connected on the active chain. Disconnecting and vault reset both clear stored permissions.
- **Signing request context** — every signing prompt shows the verified requesting site, account, request type, and connection time.
- **Log hardening** — PINs, encrypted key material, and zkProof objects are stripped before logging. `redactSensitive()` sanitizes arbitrary message payloads; a regression test fails if a sensitive field reaches a logger call.

### 3. Gated transaction signing + result ↔ request binding (#123)

Transactions were signed blind: no client-side decode, informational-only risk warnings, and signing results matched requests by `windowId` alone.

- **Parsed transaction review** — `parseTxBytes` returns a `displayValue` and a structured `reviewValue`; both are shown in the signing prompt.
- **Risk findings** — `reviewTransaction` classifies findings as danger (publish/upgrade/transfer, undecodable payload) or warning (Move calls, shared objects) and renders them above the payload.
- **Approval gate** — any danger-class finding, including an undecodable payload, now requires an explicit "I understand the risks" acknowledgement before Approve is enabled. Previously a blind/opaque payload could be approved with one click.
- **Sponsored transactions** — the sponsored flow signed BCS bytes without decoding them (`atob(sponsoredTxB64) → sign`). It now decodes through the same pipeline, shows the parsed payload + risk panel, and is gated behind the same acknowledgement.
- **Result ↔ request binding** — the background mints a per-approval `requestId` when opening the popup; a signing result must match both `windowId` and `requestId` before settling, so a stale or forged result for a reused window can't settle a different request.
- **Removed dead code** — `deriveEncryptionKey` (a weak JWT-claim key derivation with no live callers) removed so it can't be reintroduced into a key path.

### 4. Test coverage (#128)

New and updated tests across the full stack: content-script message handling, background auth and wallet handlers, dApp login and pending-auth flows, message handler routing, token-material guards, transaction risk review, signing component flows, and wallet hooks. All new security invariants (requestId mismatch, cross-origin drop, sensitive-field logging, acknowledgement gate) are directly exercised.

### 5. SECURITY.md

Adds a responsible-disclosure policy with a private reporting address (`security@fenris.com`) and guidance on AI-generated reports.

### 6. Remove `scripting` permission from manifest

The `scripting` permission was being injected by WXT but is unused. A build hook strips it so the extension doesn't request unnecessary access.


ccp-bofai (approved):
LGTM

PR #129, opened 2026-06-22
fix(security): keeper sender guard, gated logger, OAuth state

fcsondheim:
- Route keeper messages only from extension-origin senders (`isExtensionSender`, extracted to `senderGuard.ts` and shared with messageHandler)
- Replace bare console.error in keeper vaultHandlers with the gated logger
- Add a random state param to the OAuth authorize URL; verify the echoed value in both extLogin and dappLogin callbacks before exchanging the code
- Strip zkLogin salt from the User profile before writing to sessionStorage in both CallbackScreen (initial login) and persistEnrichedUser (silent renew); salt is kept in the in-memory Zustand store for signing and re-derived on demand via getUserForNetwork


ccp-bofai (approved):
LGTM

PR #134, opened 2026-06-29
feat(security): Argon2id vault KDF and web/extension CSP

fcsondheim:
Proactive hardening of the vault crypto and HTTP security posture.

### Changes
- **Argon2id key derivation** across the shared key utilities (`packages/shared/src/utils/keys`), replacing the prior derivation. Memory-hard parameters: 64 MiB, t=3, p=1, 256-bit key, random per-derivation salt. Adds `hash-wasm` (MIT, inline WASM).
- **Web vault PIN** now uses a salted Argon2id verifier instead of a plain hash. Removed an unused hashing helper (the SHA-256 used for the OAuth PKCE challenge is unchanged).
- **Extension** keeper inherits the Argon2id derivation; updated the MV3 CSP to permit the WASM module.
- **Web** now sends a Content-Security-Policy (with Trusted Types) and standard security headers (`X-Content-Type-Options`, `X-Frame-Options`, `Referrer-Policy`, HSTS, `Permissions-Policy`) via Amplify.

### Notes
- App is pre-release, so no data migration is included.

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -37,12 +37,20 @@ applications:
 # The CSP is shipped as Report-Only first (per rollout plan): it reports
 # violations without breaking the app. Once the reports are clean, rename the
 # key to `Content-Security-Policy` to enforce it.
+# TODO(tracking issue): file an issue to own the Report-Only -> enforce flip so
+# this doesn't silently live in Report-Only forever. Link it here.
 #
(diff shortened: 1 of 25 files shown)


ccp-bofai (approved):
LGTM

PR #135, opened 2026-06-29
feat: auto-lock vault on unlock-window expiry

fcsondheim:
Consolidates the vault unlock-window timing into one shared helper, then builds an always-on **auto-lock** on top of it so the wallet locks itself when the unlock window expires. Also adds a small dev-mode readout of the remaining time.

Previously each surface (web app and browser extension) tracked the unlock expiry with its own ad-hoc fields, and the window was only enforced lazily. This PR makes the timing logic single-sourced and the lock proactive.

### 1. Shared `VaultSession` (timing, single source of truth)
- New `VaultSession` helper in `packages/shared` with a minimal surface: `unlock()`, `isActive()`, `remainingMs()`, `clear()`. It owns *only* the expiry timestamp — no keys.
- `WebVaultService` (web) and the extension keeper state now both use it instead of duplicating `Date.now()`-based expiry fields, so the two surfaces can't drift.

### 2. Auto-lock
- `useVaultAutoLock` hook, mounted once at the web app root and in the extension popup. While the vault is unlocked it schedules a lock at the expiry, re-arms on unlock, and re-checks when the tab/popup regains visibility (so a timer throttled while hidden, or an expiry that passed while backgrounded, still locks promptly on return).
- **Extension keeper self-lock**: the extension holds the key in an offscreen document that outlives the popup, so the keeper now schedules its own lock at expiry. This guarantees the vault clears even when no popup is open.
- Web is fully covered by the hook (the vault lives in the page context; closing the tab discards it).

### 3. Dev-mode unlock countdown (display only)
- `useUnlockTimeRemaining` hook + a dev-mode-only row in the header dropdown showing the remaining time (`m:ss`). Web reads the in-process session; the extension queries the keeper once per popup open and counts down locally. Display only.

Screenshot 2026-06-30 at 12 01 32

### 4. Supporting plumbing
- Added a `GET_UNLOCK_REMAINING` vault/keeper message and `getUnlockRemainingMs()` on the vault services so the UI can read remaining time across the extension's message boundary.


ccp-bofai (approved):
LGTM

PR #138, opened 2026-07-02
feat: remove enoki api from `getZkLoginAddress`

fcsondheim:
Moves the zkLogin address lookup off Enoki (`api.enoki.mystenlabs.com/v1/zklogin`) onto the EVE Frontier auth API (`/auth/zklogin`), hardens the surrounding error handling, and removes the Enoki API key from that path.

### Changes

**Address lookup**
- `getZkLoginAddress`: call `/auth/zklogin` with the host/tenant derived from the JWT via `getApiContext`; throw with the real status + body on failure; validate `salt`/`address` and cache only successful responses.
- Drop `enokiApiKey` from `GetZkLoginAddressParams` and remove the `ZkLoginAddressResponse` (Enoki envelope) type.
- Update callers (`userJwtSync`, `CallbackScreen`, `dappLogin`, `authStoreUtils`) to the throw-on-failure contract.
- Remove the now-unused `getEnokiApiKey` helper and its re-exports.

**Docs / config**
- Remove `VITE_ENOKI_API_KEY` from `.envsample` and `README.md`.
- Update `docs/TROUBLESHOOTING.md` address-derivation guidance to reference the JWT / API Gateway instead of the Enoki key.

**Tests**
- Update affected `shared` and extension tests for the new endpoint, error contract, and single-arg `enrichUserWithZkLoginIfNeeded`.

### DIFF
--- a/packages/shared/src/auth/getZkLoginAddress.ts
+++ b/packages/shared/src/auth/getZkLoginAddress.ts
@@ -2,11 +2,9 @@ import type { ZkLoginAddressData } from '#/types/enoki'
 import { getApiContext } from './getApiContext'
 import type { GetZkLoginAddressParams } from './types'
 
+// Keyed by JWT: the token uniquely identifies the user/session the address
+// is derived for.
 const cache = new Map<string, ZkLoginAddressData>()
-
-function cacheKey(params: GetZkLoginAddressParams): string {
-  return `${params.jwt}`
-}
 
 /**
  * Clear the in-memory cache of zkLogin address lookups.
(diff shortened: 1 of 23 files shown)


packages/shared/src/auth/getZkLoginAddress.ts
```diff
@@ -1,10 +1,11 @@
...
+import type { ZkLoginAddressData } from '#/types/enoki'
+import { getApiContext } from './getApiContext'
 import type { GetZkLoginAddressParams } from './types'
 
-const cache = new Map<string, ZkLoginAddressResponse>()
+const cache = new Map<string, ZkLoginAddressData>()
 
 function cacheKey(params: GetZkLoginAddressParams): string {
-  return `${params.enokiApiKey}:${params.jwt}`
+  return `${params.jwt}`
```
ccp-bofai: Wondering if `params.jwt` is simply a string? and that returning it might result in the same result as this?

ccp-bofai (approved):
One suggestion to check but LGTM

PR #142, opened 2026-07-08
 fix: verify JWT signatures against tenant JWKS instead of trusting decoded claims

fcsondheim:
Adds signature verification for a JWT at every point a JWT first enters client trust:

- **`verifyIdTokenForTenant`** (new): verifies an id_token's signature against the JWKS of the FusionAuth instance a login was actually initiated against — resolved from tenant config, never inferred from the token's own claims, to avoid the same trust-circularity as a prior tier/host-injection issue.
- **Extension token exchange**: `exchangeCodeForToken` now verifies the returned id_token and checks its `nonce` claim against the nonce originally sent — previously there was no validation of any kind here (unlike the web flow, which at least checks nonce consistency via oidc-client-ts).
- **Web login/renewal**: the OAuth redirect callback and silent-renew path both verify the fresh id_token from FusionAuth before trusting it.
- **Vended zkLogin JWT**: the token minted by the app's own backend for zkLogin proofs is verified against the same per-tenant FusionAuth JWKS before use.
- **Issuer discovery fix**: verification initially failed in practice because FusionAuth's actual `iss` claim (discovered from `.well-known/openid-configuration`) differs from the authority URL used for OAuth endpoints. The issuer is now discovered rather than assumed.

All failure modes (bad signature, wrong issuer/audience, expired token, unreachable JWKS/discovery endpoint) fail closed. Verification failures propagate as hard auth failures, never silently fall back to trusting the token.


packages/shared/src/auth/verifyJwt.ts
```diff
@@ -0,0 +1,58 @@
...
+      .then((response) => response.json())
+      .then((config: { issuer: string }) => config.issuer)
+      .catch((error) => {
+        issuerCache.delete(normalized)
+        throw error
+      })
+    issuerCache.set(normalized, issuer)
+  }
+  return issuer
+}
```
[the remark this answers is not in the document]
fcsondheim: Confirmed — jose skips the entire issuer check when `issuer` is falsy/`undefined` (`jwt_claims_set.js`), so a bad discovery response would've silently disabled issuer validation. Fixed: `getIssuerForServer` validates `response.ok` and that `issuer` is a non-empty string before caching, throwing otherwise. Added regression tests for both cases.

packages/shared/src/auth/__tests__/verifyJwt.test.ts
```diff
@@ -0,0 +1,227 @@
...
+let testCounter = 0
+
+const signToken = async (
+  overrides: {
+    issuer?: string
+    audience?: string
+    expiresIn?: string
+    kid?: string
+    key?: CryptoKey
+    claims?: Record<string, unknown>
```
ccp-bofai: Can we please use standard claim names so it is clear and obvious as to what is being set. i.e. https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3 for `aud`

packages/shared/src/auth/__tests__/verifyJwt.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  it('rejects a token with the wrong issuer', async () => {
+    const token = await signToken({ issuer: 'https://attacker.example.com' })
+    await expect(
+      verifyIdTokenForTenant(token, 'stillness' as never),
+    ).rejects.toThrow()
+  })
+
+  it('rejects a token with the wrong audience', async () => {
+    const token = await signToken({ audience: 'someone-elses-client' })
```
ccp-bofai: Looking at this I don't see this as an `aud` claim but a claim called `audience` which should probably be irrelevant, which makes this test not as clear as it should be to read.

packages/shared/src/auth/__tests__/verifyJwt.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  it('rejects a token signed with a key not in the JWKS', async () => {
+    const { privateKey: otherKey } = await generateKeyPair('RS256')
+    const token = await signToken({ key: otherKey })
+    await expect(
+      verifyIdTokenForTenant(token, 'stillness' as never),
+    ).rejects.toThrow()
+  })
+
+  it('rejects a token with the wrong issuer', async () => {
+    const token = await signToken({ issuer: 'https://attacker.example.com' })
```
ccp-bofai: `iss` claim maybe, why not call it that?

packages/shared/src/auth/__tests__/verifyJwt.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  it('rejects a token with the wrong audience', async () => {
+    const token = await signToken({ audience: 'someone-elses-client' })
+    await expect(
+      verifyIdTokenForTenant(token, 'stillness' as never),
+    ).rejects.toThrow()
+  })
+
+  it('rejects an expired token', async () => {
+    const token = await signToken({ expiresIn: '-1h' })
```
ccp-bofai: is this an `exp` claim in disguise?

packages/shared/src/auth/verifyJwt.ts
```diff
@@ -0,0 +1,71 @@
...
+  idToken: string,
+  tenantId: TenantId,
+): Promise<JWTPayload> {
+  const { serverUrl, clientId } = getTenantConfig(tenantId)
+  const jwks = getJwksForServer(serverUrl)
+  const issuer = await getIssuerForServer(serverUrl)
+  const { payload } = await jwtVerify(idToken, jwks, {
+    issuer,
+    audience: clientId,
+  })
```
ccp-bofai, line 69: This probably breaks when the keys on the JWKS get rotated. Have to be careful here as this is also a ddos vector too potentially. Essentially when you see a `kid` that is not in the cache you probably need to invalidate the cache and refetch from the JWKS. Bad actors could potentially send JWTs your way with random `kid` values to continually cause the cache to be invalidated and to constantly put load on the JWKS (throttling is probably an idea, but maybe there are better ways of dealing with this issue)

ccp-bofai (changes requested):
I think this will break when the JWKS rotates keys. Comments where I see the issue happening.

packages/shared/src/auth/verifyJwt.ts
fcsondheim: This behaviour is handled by the `jose` library: https://github.com/panva/jose/blob/main/docs/jwks/remote/functions/createRemoteJWKSet.md

> The JSON Web Key Set is fetched when no key matches the selection process but only as frequently as the cooldownDuration option allows to prevent abuse.

ccp-bofai (approved):
LGTM

PR #143, opened 2026-07-09
refactor: use address aliasing from `wallet-core`

fcsondheim:
Address aliasing is now core zklogin account functionality in `@evefrontier/wallet-core` (see [wallet-core#44](https://github.com/evefrontier/wallet-core/pull/44)), so the chain-facing layer added in #137/#139 moves there and this repo keeps only the React/UI layer.

- Deletes `useAddressAliases.{config,transaction,validation,helpers}.ts` in favor of imports from `@evefrontier/wallet-core/address-alias`
- `useAddressAliases.query.ts` keeps only the react-query wrapper; `useAddressAliases` adapts to the new `executeAddressAliasTx` signature (explicit `sender` + a `signTransaction` adapter around the signing-context callback)
- `transactionRiskReview.ts` uses the shared `isAddressAliasCall` instead of its local copy
- No behavior change: same PTBs, validation messages, hook API, and risk finding; existing tests pass unchanged


ccp-bofai (approved):
LGTM but I think its probably an idea to look into the comment from CoPilot as that is possibly a bit of a  potential pain for some builders that import that function

PR #145, opened 2026-07-14
feat(web): add desktop sidebar with centralised `/wallet` auth guard

fcsondheim:
## Desktop sidebar and wallet navigation

Adds the desktop sidebar to the web app and cleans up the routing that supports it.

### Sidebar
- New left sidebar for the web app, shown above the mobile breakpoint (tablet/desktop), with **Home**, **Transactions**, and **Address Aliases** entries.
- Nav items now point at the real route paths (`/wallet`, `/wallet/transactions`, `/wallet/address-aliases`); the stale `ROUTE_PATHS` constant and dead `NavPath` union members were removed.

Screenshot 2026-07-14 at 13 27 20

### Routing
- Centralised the auth guard into a `/wallet` layout route: `beforeLoad` runs `requireAuth` once for the whole subtree instead of being repeated in each child route, so any future `/wallet` route is guarded automatically.
- Navigating to a wallet route while signed out (e.g. via a sidebar link) redirects to sign-in — or the PIN unlock screen when the vault is locked — instead of rendering an error.

### Fixes
- zkLogin proof requests now include the selected network; `network` is required on `ZkProofParams` so call sites can't omit it (with test coverage for the request body).
- Extension wallet view surfaces errors as toasts instead of inline text.

### DIFF
--- a/apps/extension/src/features/wallet/components/AuthenticatedWalletView.tsx
+++ b/apps/extension/src/features/wallet/components/AuthenticatedWalletView.tsx
@@ -5,12 +5,14 @@ import {
   TenantSelector,
   Text,
   TokenListSection,
+  useToast,
 } from '@evevault/shared/components'
 import { useUnlockTimeRemaining } from '@evevault/shared/hooks'
 import { EXTENSION_ROUTES, getSuiscanUrl } from '@evevault/shared/utils'
 import type { SuiChain } from '@mysten/wallet-standard'
 import { useNavigate } from '@tanstack/react-router'
 import type { User } from 'oidc-client-ts'
+import { useEffect } from 'react'
 import { APP_VERSION } from '@/lib/appVersion'
 
 /**
@@ -78,11 +80,21 @@ export function AuthenticatedWalletView({
   onNetworkSwitchStart: (previousNetwork: string, targetNetwork: string) => void
 }) {
   const navigate = useNavigate()
+  const { showErrorToast } = useToast()
   const openFaucet = faucetUrl
     ? () => window.open(faucetUrl, '_blank', 'noopener,noreferrer')
     : undefined
   // Display only of the vault unlock window.
   const unlockRemainingLabel = useUnlockTimeRemaining(devMode) ?? undefined
+
+  // Surface auth/device failures as transient toasts rather than inline text.
+  useEffect(() => {
+    if (authError) showErrorToast('AuthError', authError)
+  }, [authError, showErrorToast])
+
+  useEffect(() => {
+    if (deviceError) showErrorToast('DeviceError', deviceError)
+  }, [deviceError, showErrorToast])
 
   return (
     <div className="flex flex-col h-full">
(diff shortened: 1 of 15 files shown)


apps/extension/src/features/wallet/components/AuthenticatedWalletView.tsx
```diff
@@ -78,12 +80,22 @@ export function AuthenticatedWalletView({
...
+  const { showErrorToast } = useToast()
   const openFaucet = faucetUrl
     ? () => window.open(faucetUrl, '_blank', 'noopener,noreferrer')
     : undefined
   // Display only of the vault unlock window.
   const unlockRemainingLabel = useUnlockTimeRemaining(devMode) ?? undefined
 
+  // Surface auth/device failures as transient toasts rather than inline text.
+  useEffect(() => {
+    if (authError) showErrorToast('AuthError', authError)
```
OliValur, line 92: Since this is surfaced to the user this might be a bit techy of an error message I feel, would it be a good fit to have something more human readable here?

apps/extension/src/features/wallet/components/AuthenticatedWalletView.tsx
```diff
@@ -78,12 +80,22 @@ export function AuthenticatedWalletView({
...
 }) {
   const navigate = useNavigate()
+  const { showErrorToast } = useToast()
   const openFaucet = faucetUrl
     ? () => window.open(faucetUrl, '_blank', 'noopener,noreferrer')
     : undefined
   // Display only of the vault unlock window.
   const unlockRemainingLabel = useUnlockTimeRemaining(devMode) ?? undefined
 
+  // Surface auth/device failures as transient toasts rather than inline text.
```
OliValur, line 90: Are there some errors that might warrant being persisted inline while some are more toasty? Cursor spit this out for which could be which, not blocking just something I thought might cause user uncertainty

"Toast is fine

One-off action failures that the screen already handles (send, aliases) — those flows often toast already
Stale/authError leftovers that don’t actually break the current session
Soft “try again” cases where browsing still works
Prefer persistent UI

ZK / signing broken while unlocked (deviceError from proof gen) — toast vanishes, wallet looks healthy, send still fails. Keep a banner or disable send with an explanation until cleared
PIN / unlock / init — already right on LockScreen (don’t move those to toast-only)
Hard init — already right on InitErrorView + Reload"

OliValur (approved):

apps/extension/src/features/wallet/components/AuthenticatedWalletView.tsx
fcsondheim: I agree that human readable would be nice but at this point I'm wanting to maintain the same surfacing that already happens without changing any of the behaviour :) Since we will be doing quite an extensive UI/UX review this doesn't need to be optimised for UX just yet

apps/extension/src/features/wallet/components/AuthenticatedWalletView.tsx
fcsondheim: Nice call! As Cursor mentioned some of the errors are already persisted inline. However, much like the previous comment we will be doing a bigger review overall once we do the UI/UX overhaul and we can take a look at the remaining error surfacing then and how they are communicated to the user.

PR #146, opened 2026-07-15
feat: reactive recovery for epoch expiry

fcsondheim:
zkLogin proofs live at most one epoch, so the fullnode can reject a signature with `ZKLogin expired at epoch N`. Previously this surfaced as a failed transaction and required signing out and back in.

- New `withZkLoginEpochRetry`: detects the expiry rejection, rotates the ephemeral key (fresh nonce/JWT/proof), and re-signs the same transaction bytes once. Rotation doesn't change the zkLogin address, so the bytes stay valid. Other errors propagate; localnet is skipped.
- New shared `signAndExecuteTransaction` wrapping sign→execute inside the retry; replaces the duplicated inline flow in `useSendToken`, `useDevMode`, and the WalletScreen test transaction. Throws typed `TransactionFailedError` on execution failure (never retried).
- Removed dead `chain ?? SUI_TESTNET_CHAIN` fallbacks in WalletScreen — the context store always provides a defined chain, so the retry now trivially targets the actual submission network.

### DIFF
--- a/packages/shared/src/wallet/signAndExecute.ts
+++ b/packages/shared/src/wallet/signAndExecute.ts
@@ -30,7 +30,8 @@ type SignTransactionFn = (
  * second expiry rejection, propagates to the caller.
  *
  * Throws TransactionFailedError when the chain executes the transaction but
- * reports it as failed (this is never retried).
+ * reports it as failed, or when a successful response is missing a digest
+ * (this is never retried).
  */
 export const signAndExecuteTransaction = async ({
   chain,
@@ -42,7 +43,7 @@ export const signAndExecuteTransaction = async ({
   suiClient: SuiGrpcClient
   txBytes: Uint8Array
   sign: SignTransactionFn
-}): Promise<string | null> =>
+}): Promise<string> =>
   withZkLoginEpochRetry(chain, async () => {
     const { signature } = await sign('TransactionData', txBytes)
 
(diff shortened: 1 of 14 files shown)


packages/shared/src/wallet/signAndExecute.ts
```diff
@@ -0,0 +1,72 @@
...
+type SignTransactionFn = (
+  scope: 'TransactionData',
+  bytes: Uint8Array,
+) => Promise<{ bytes: string; signature: string }>
+
+/**
+ * Signs pre-built transaction bytes and submits them, returning the digest.
+ *
+ * Handles zkLogin proof expiry transparently: proofs are valid for at most one
+ * epoch (see withZkLoginEpochRetry), so the fullnode may reject the signature
```
vayan, line 25: Should we then get proofs that are valid for current epoch + 1? 

That way we can refresh in the background when you detect an outdated epoch but at least everything would still work without retries.

vayan (approved):
Just one question but looks good!

packages/shared/src/wallet/signAndExecute.ts
fcsondheim: Prefer not to, that goes against the intended design - bounding it to the current epoch is meant to minimise any risk that could occur with leaked zkproofs. Limiting the zkproof to maxepoch is an intentional security boundary.

If latency becomes an issue such that we need to look at proactive refreshing, we can consider that in the future but without overtly extending maxEpoch.

PR #147, opened 2026-07-15
fix: persist vault unlock window across page loads

fcsondheim:
Creating a PIN starts the 10-minute unlock window, but the window lived only in page memory — the OAuth sign-in redirect (a full page load) wiped it, so users were re-prompted for their PIN seconds after setting it, and again on later redirects.

- `VaultSession` now accepts an optional persistence adapter; the web vault backs it with per-tab `sessionStorage` and, on load, restores the signer from IndexedDB when the persisted window is still open. Remaining time is preserved, never extended; invalid or oversized expiry values are rejected.
- Device-store rehydration now refreshes lock state on web as well as extension (`refreshVaultLockState`), and falls back to locked if the refresh fails.
- Security posture is unchanged: the PIN verifier is a UX-level presence check (the keypair handle already lives in IndexedDB; the boundary is the non-extractable CryptoKey). A new tab or browser restart still prompts.

**Also in this PR**

- Renamed a `VaultSession` constructor param from `storage` to `persistence`: WXT's auto-import scanner treats bare `storage` identifiers as its `wxt/utils/storage` global and injected an unresolvable import into the extension bundle.
- Test-env fix for Node 25, which ships a non-functional `localStorage`/`sessionStorage` global (without `--localstorage-file`) that shadows jsdom's: `vitest.setup.ts` replaces broken storage globals with an in-memory implementation. This un-breaks 14 pre-existing jsdom test failures under the repo's pinned Node version.


OliValur (approved):
👏

PR #150, opened 2026-08-05
refactor: use wallet-core modules

fcsondheim:
Bumps `@evefrontier/wallet-core` from `0.0.5` → `0.0.6` and replaces locally duplicated logic with the equivalent wallet-core modules, so URL formation, JWT decoding, and transaction parsing live in one shared place.

## Changes
- **jwt module**
  - `getApiContext` now delegates to `@evefrontier/wallet-core/jwt`, passing an `apiHostTemplate` instead of hardcoding tier/tenant resolution locally. Return field `decoded` → `claims`.
  - `decodeJwtSafely` is now a thin wrapper over `decodeJwtPayloadSafely`.
- **transaction module**
  - Re-export `parseTransactionBytes` / `ParseTransactionBytesResult` from wallet-core; deleted the local `parseTxBytes.ts` and its node test (~390 lines).
- **sponsored-transaction module**
  - `sponsoredTransactionHandler` uses `fetchUnsignedSponsoredTransaction` and `executeSponsoredTransaction` in place of hand-rolled `fetch` calls, response validation, and the local `SponsoredTxReturn` type. A small `sponsoredApiContext` helper builds the wallet-core API context (`pathPrefix` handles the prepare endpoint living under `/v2` while execute sits at the gateway root).
- Dependency bumps: `@mysten/sui` `^2.22.1` → `^2.23.1`, plus dev-dep bumps (playwright, turbo). Regenerated `routeTree.gen.ts` and `bun.lock`.


apps/extension/src/lib/background/handlers/sponsoredTransactionHandler.ts
```diff
@@ -145,45 +140,19 @@ async function handleSponsoredTransaction(
...
-          url: metadata?.url,
-        }),
-        headers: {
-          'X-Tenant': tenant,
-          'Content-Type': 'application/json',
-          Authorization: `Bearer ${jwt.id_token}`,
-        },
+        txAction: action,
+        // repo forwards assemblyId as a string; wallet-core types it as number
+        assembly: assembly as unknown as number,
```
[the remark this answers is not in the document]
fcsondheim: sending a string here is intentional and correct. The v2 sponsored endpoints take `assemblyId` as a uint64 decimal *string* specifically to avoid JS number precision loss for ids > 2^53. The cast temporarily fixes a type-level mismatch which is fixed in https://github.com/evefrontier/wallet-core/pull/52

apps/extension/src/lib/background/handlers/sponsoredTransactionHandler.ts
```diff
@@ -57,36 +74,14 @@ async function executeSponsoredTx({
...
-        }),
-        headers: {
-          'X-Tenant': tenant,
-          'Content-Type': 'application/json',
-          Authorization: `Bearer ${idToken}`,
-        },
-      },
+    const { digest, effects } = await executeSponsoredTransaction(
+      { preparationId, userSignatureB64Bytes: zkSignature },
+      sponsoredApiContext(apiBaseUrl, tenant, idToken),
```
ccp-bofai, line 80: Thought we were going to use the tenant from the token, rather than passing the tenant around? or is this something that happens in later cleanup or am I missing something?

packages/shared/src/auth/utils/jwtUtils.ts
```diff
@@ -1,13 +1,7 @@
-import { decodeJwt, type JWTPayload } from 'jose'
+import { decodeJwtPayloadSafely } from '@evefrontier/wallet-core/jwt'
+import type { JWTPayload } from 'jose'
```
ccp-bofai, line 2: Is removing 'jose' a journey that will continue from wallet-core  to evevault or is this a leftover?

ccp-bofai (commented):
Couple of comments

packages/shared/src/auth/utils/jwtUtils.ts
fcsondheim: I hadn't planned on removing `jose` from evevault :)

apps/extension/src/lib/background/handlers/sponsoredTransactionHandler.ts
fcsondheim: This actually happens earlier - the tenant is decoded from the jwt claims in the _jwt module_, then forwarded as-is into the sponsored transaction module. The sponsored transaction module is transport-only and doesn't perform any of that decoding in-module.

ccp-bofai (approved):
LGTM

PR #151, opened 2026-08-06
refactor: zkLogin epoch timing with wallet-core epoch module

fcsondheim:
Routes both zkLogin epoch fetchers through `@evefrontier/wallet-core/epoch` instead of hand-rolling epoch/`maxEpoch` timing math in each fetcher. Epoch semantics now live in one shared place.

`getCurrentEpochFromGraphQL` and `getCurrentEpochFromRpc` keep their existing return shape (`{ numericMaxEpoch, maxEpochTimestampMs }`), so every downstream consumer is untouched.

## Changes

- **`graphqlEpoch.ts`** — builds a `ChainEpochInfo` from the GraphQL response and computes timing with `computeEpochState(info, { epochsFromCurrent: 0, ... })`. Behavior-preserving: with `epochsFromCurrent: 0`, `maxEpochTimestampMs` resolves to the same value as before in every branch (real end timestamp when present, `start + 24h` fallback otherwise).
- **`rpcEpoch.ts`** — replaces `ledgerService.getEpoch()` with `fetchEpochFromSystemState(client)`, which reads the on-chain system-state object. This sources the **actual** `epochDurationMs` from chain parameters rather than always assuming the 24h fallback, so localnet's `maxEpochTimestampMs` now reflects the chain's real epoch length. Non-numeric system-state fields now throw instead of silently coercing to `0`.

- Centralizes duplicated timing math and removes the local `DEFAULT_EPOCH_DURATION_MS` reliance on the gRPC path.

### DIFF
--- a/packages/shared/src/sui/__tests__/rpcEpoch.test.ts
+++ b/packages/shared/src/sui/__tests__/rpcEpoch.test.ts
@@ -2,10 +2,10 @@ import { SUI_LOCALNET_CHAIN } from '@mysten/wallet-standard'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import { DEFAULT_EPOCH_DURATION_MS } from '#/utils/constants'
 
-const mockGetEpoch = vi.fn()
+const mockGetCurrentSystemState = vi.fn()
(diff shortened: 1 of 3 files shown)


ccp-bofai (approved):
LGTM

PR #152, opened 2026-08-06
refactor(extension): adopt WXT browser API + enable Firefox dev

fcsondheim:
Groundwork for Firefox support. Moves `apps/extension` off the Chrome-only `chrome.*` global onto WXT's cross-browser `browser` API, makes the test suite browser-native, and wires up a Firefox dev workflow — so the extension compiles, builds, tests, and dev-runs on **both Chrome and Firefox**.

Behaviour-preserving on Chrome. This does **not** yet make Firefox functional end-to-end.

## Changes

**Browser API migration**
- Replace all `chrome.*` value calls with `browser.*` (`wxt/browser`) and the `MessageSender` / `Port` / `storage.StorageChange` types with `Browser.*`; remove the `/// <reference types="chrome">` directives.
- Convert callback-style APIs to promises/`await` — required because Firefox's `browser` is promise-only: `sendToKeeper` retry loop, storage reads, and `launchWebAuthFlow` in ext/dapp login (rejections now route through `sendAuthError`); `lastError` checks become `try/catch`.
- Adjust for stricter webextension-polyfill types: shape-cast `StorageChange.newValue`, guard `windows.create`'s optional result.

**Browser-native tests**
- Alias `wxt/browser` to a live proxy over `globalThis.browser` (`vitest.browser-mock.ts`) so per-test `vi.stubGlobal('browser', …)` works despite the module's import-time snapshot.
- Convert all mocks/assertions from `chrome` to `browser`; drop the now-unused `@types/chrome` dependency.

**Firefox build & dev**
- Add `browser_specific_settings.gecko.id` (Firefox-only, via manifest hook) — Firefox MV3 won't load an extension without it.
- Fix `webExt` config to use WXT's real `binaries` map (the old `chromiumExecutablePath` / `executablePath` keys were silently ignored).
- `bun run dev` now runs web + Chrome ext + Firefox ext together (`turbo run dev dev:firefox`); ports: Chrome 3000, Firefox 3002 (`--port`), web 3005. Added `dev:chrome` (Chrome-only) and `dev:ext:firefox`.

### DIFF
--- a/apps/extension/src/lib/background/handlers/auth/authHelpers.ts
+++ b/apps/extension/src/lib/background/handlers/auth/authHelpers.ts
@@ -89,7 +89,9 @@ export function sendExtensionAuthSuccess(id: string, jwt: JwtResponse): void {
     type: AuthMessageTypes.AUTH_SUCCESS,
     token,
   }
-  browser.runtime.sendMessage(message)
+  // Fire-and-forget to the popup; browser.runtime.sendMessage rejects when no
+  // extension page is listening (e.g. popup closed), which is expected here.
+  void browser.runtime.sendMessage(message).catch(() => {})
 }
 
 export function sendDappConnectSuccessToTab(
@@ -116,11 +118,14 @@ export function sendDappConnectSuccessToTab(
 }
 
 export function sendAuthError(id: string, error: unknown): void {
-  browser.runtime.sendMessage({
-    id,
-    type: 'auth_error',
-    error,
-  })
+  // Fire-and-forget to the popup; ignore rejection when nothing is listening.
+  void browser.runtime
+    .sendMessage({
+      id,
+      type: 'auth_error',
+      error,
+    })
+    .catch(() => {})
 }
 
 export function extractEmailFromJwt(jwt: JwtResponse): string {
--- a/apps/extension/src/lib/background/handlers/auth/authHelpers.ts
+++ b/apps/extension/src/lib/background/handlers/auth/authHelpers.ts
@@ -10,6 +10,7 @@ import { CONTEXT_STORAGE_KEY, createLogger } from '@evevault/shared/utils'
 import type { SuiChain } from '@mysten/wallet-standard'
 import { decodeJwt } from 'jose'
 import type { IdTokenClaims } from 'oidc-client-ts'
+import { browser } from 'wxt/browser'
 import { sendToTab } from '@/lib/background/messaging/tabMessaging'
 import type { MessageWithId } from '@/lib/background/types'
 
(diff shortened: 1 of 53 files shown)


apps/extension/src/lib/background/handlers/auth/authHelpers.ts
```diff
@@ -93,7 +89,9 @@ export function sendExtensionAuthSuccess(id: string, jwt: JwtResponse): void {
     type: AuthMessageTypes.AUTH_SUCCESS,
     token,
   }
-  chrome.runtime.sendMessage(message)
+  // Fire-and-forget to the popup; browser.runtime.sendMessage rejects when no
```
ccp-bofai, line 92: if `browser.runtime.sendMessage rejects` then shouldn't this function be `async` and line 94 changed so that `sendMessage` is awaited inside a try/catch.

apps/extension/src/lib/background/handlers/auth/authHelpers.ts
```diff
@@ -120,11 +118,14 @@ export function sendDappConnectSuccessToTab(
...
-    error,
-  })
+  // Fire-and-forget to the popup; ignore rejection when nothing is listening.
+  void browser.runtime
+    .sendMessage({
+      id,
+      type: 'auth_error',
+      error,
+    })
+    .catch(() => {})
```
ccp-bofai, line 128: Promise's in a none async function

apps/extension/src/lib/background/handlers/auth/dappLogin.ts
```diff
@@ -532,25 +535,25 @@ export async function handleDappLogin(
...
+    codeVerifier,
+    state,
+    nonce,
+    tabId,
+    dappContext,
+  }
+  void browser.identity
+    .launchWebAuthFlow({ url: authUrl.toString(), interactive: true })
+    .then((responseUrl) => handleOAuthCallback(responseUrl, ctx))
+    .catch((error) => handleOAuthCallback(undefined, ctx, error))
```
ccp-bofai, line 558: this is inside an async function where await is being used. This could do the same inside its own try/catch.

apps/extension/src/lib/background/handlers/auth/extLogin.ts
```diff
@@ -366,6 +362,8 @@ function launchOAuthLogin({
         currentChain,
         tenantId,
       })
-    },
-  )
+    })
+    .catch((error) => {
+      sendAuthError(id, error)
+    })
 }
```
ccp-bofai, line 369: Probably should be marked async

apps/extension/src/lib/background/handlers/messageHandler.ts
```diff
@@ -301,7 +302,7 @@ function hasRequiredSender(sender: MsgSender, access: RouteAccess): boolean {
  */
 function broadcastChangeEvent(payload: unknown): void {
   log.info('Broadcasting chain change event', payload)
-  chrome.tabs.query({}, (tabs) => {
+  void browser.tabs.query({}).then((tabs) => {
```
ccp-bofai, line 305: Promise inside none async function

ccp-bofai (approved):
Looks good but I think a follow up PR might be an idea. Seems to be some promise based code in none async functions.

fcsondheim:
agree about the async, out of scope for this PR but will do a cleaner sweep separately

PR #153, opened 2026-08-07
refactor(shared): browser-agnostic API sweep (chrome.* → browser.*)

fcsondheim:
`chrome.*` → WXT `browser.*` namespace sweep applied to PR #152's `apps/extension`, now for `packages/shared`.

Mechanical namespace migration only — no behavior change on Chrome.

## Why

`packages/shared` is consumed by both the extension and the web app. The extension APIs it touched went through the ambient `chrome` global with callback-style calls, which Firefox's promise-only `browser` API doesn't support. This routes them through `@wxt-dev/browser` and converts callbacks to `await`, keeping Chrome working while unblocking Firefox.

## Changes

- **Dependency:** add `@wxt-dev/browser` to `packages/shared`; source imports `{ browser }` from it (not `wxt/browser` — shared isn't a WXT project).
- **Namespace sweep** across 15 source files: `chrome.*` → `browser.*`; existence guards `typeof chrome === 'undefined' || !chrome.storage` → `!browser?.storage`; `isExtension()` now gates on `browser?.runtime?.id`.
- **Callback → promise** conversions (the Firefox-breakers): storage `get`/`set`/`remove` and `identity.launchWebAuthFlow` in `adapters/extension`, `authCleanup`, `resetVaultOnDevice`, `initPersistenceHelpers`, `authConfig`, `authExtensionChrome`.
- **Type shim removed:** delete the local `src/types/browser.ts` (`Chrome*` interfaces) and the ambient `chrome` global in `globals.d.ts` (kept the `NodeJS.Timeout` shim); types now come from `@wxt-dev/browser`.
- **Tests → browser-native:** stubs moved from `globalThis.chrome` → `globalThis.browser`; callback mocks are now promise-returning; a live-proxy mock (`vitest.browser-mock.ts`) is aliased per test project so per-test `browser` stubs are visible despite the import-time snapshot.
- Dual web/extension guards (`isWeb`/`isExtension`) preserved throughout.


packages/shared/src/utils/getters.ts
```diff
@@ -20,7 +21,7 @@ export const getDeviceData = async (chain: SuiChain) => {
   }
 
   // Fallback: read from storage only if store is missing data
-  const result = await chrome.storage.local.get([DEVICE_STORAGE_KEY])
+  const result = await browser.storage.local.get([DEVICE_STORAGE_KEY])
   const persistedState = parsePersistedState(result[DEVICE_STORAGE_KEY])
```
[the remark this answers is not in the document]
fcsondheim: This is dead code and `getDeviceData` is only called in its own tests. Removing in PR #156

packages/shared/src/auth/authConfig.ts
```diff
@@ -170,21 +170,19 @@ export function redirectToFusionAuthLogout(): void {
...
-    isExtension() &&
-    typeof chrome !== 'undefined' &&
-    chrome.identity?.launchWebAuthFlow
-  ) {
-    chrome.identity.launchWebAuthFlow(
-      { url: logoutUrl.toString(), interactive: true },
-      () => {
-        chrome.runtime?.sendMessage?.({
+  if (isExtension() && browser.identity?.launchWebAuthFlow) {
+    void browser.identity
```
ccp-bofai, line 174: function is not marked async and uses thenables.

packages/shared/src/services/keeperService.ts
```diff
@@ -46,7 +47,7 @@ export const ephKeyService = {
 
     log.debug('[ephKeyService] Unlocking vault')
 
-    const res = (await chrome.runtime?.sendMessage?.({
+    const res = (await browser.runtime?.sendMessage?.({
```
ccp-bofai, line 50: The exported function object this lives in is not marked as async (as far as I can tell from GitHub's web view).

packages/shared/src/stores/contextStore.switching.ts
```diff
@@ -173,10 +174,13 @@ const needsNetworkDataInitialization = (
...
-      payload: { chains: [chain] },
-    })
+    // Fire-and-forget; swallow rejection (no background listener).
+    void browser.runtime
+      ?.sendMessage?.({
+        __from: 'Eve Vault',
+        event: 'change',
+        payload: { chains: [chain] },
+      })
+      ?.catch(() => {})
```
ccp-bofai, line 184: The function object this lives in is not marked as async

ccp-bofai (approved):
Looks good but like in 152 there seems to be code using promises inside function objects that are not marked async. So probably something to look at in a follow up PR.

packages/shared/src/services/keeperService.ts
fcsondheim: this is marked as async in L37 - not changed in this PR, which is probably why it wasn't obvious in web view :)

fcsondheim:
same as previous PR - the async handling will be standardised in another PR to prevent scope creep in this one :)

PR #155, opened 2026-08-07
refactor(extension): extract KeeperHost seam

fcsondheim:
Introduces a `KeeperHost` interface and moves the background↔keeper transport behind `ChromeOffscreenKeeperHost`. **Pure relocation** — the Chrome offscreen path is behaviourally unchanged.

## Why

The keeper (which holds the ephemeral key in RAM) runs in a Chrome **offscreen document**, an API Firefox doesn't have. Today that dependency is spread across `offscreenService.ts` and `vaultHandlers.sendToKeeper`. This PR collects it behind one seam so a later, security-reviewed PR can add a `FirefoxKeeperHost` (background-hosted keeper) additively, without touching the handlers. It's the architectural de-risk before Firefox needs the keeper.

## Changes

- **New `src/lib/background/keeper/`:**
  - `keeperHost.ts` — `KeeperHost` interface (`ensureReady`, `send`) + the exported `keeperHost` singleton (hardcoded to the Chrome impl).
  - `chromeOffscreenKeeperHost.ts` — offscreen lifecycle (`hasDocument`/`createDocument`), the `KEEPER_READY` handshake + timeout, and the `port closed` backoff retry, moved verbatim.
  - `__tests__/chromeOffscreenKeeperHost.test.ts` — target-tagging, retry behavior, create/reuse, and readiness-wait.
- **Rewired to the seam:** `sendToKeeper` is now a one-line delegate to `keeperHost.send` (so handlers/`localVaultHandlers`/`dappLogin` are untouched); `checkKeeperUnlocked` uses `keeperHost.send(..., 1)`; `background.ts` startup warm-up uses `keeperHost.ensureReady(false)`.
- **Deleted** `services/offscreenService.ts`.

## Note

One intentional deviation: the `KEEPER_READY` listener now registers **lazily** (just before `createDocument`) instead of at module load. This removes the host's import-time side effect (which previously forced tests to mock the whole module) and is runtime-equivalent — the keeper only emits `KEEPER_READY` after its document is created. Otherwise the offscreen path is unchanged.


apps/extension/src/lib/background/handlers/__tests__/walletHandlers.test.ts
ccp-bofai: I think an async arrow function with a body that just throws the error might be equivalent and easier to understand

apps/extension/src/lib/background/handlers/auth/__tests__/dappLogin.test.ts
ccp-bofai: Just a note.... comparing this to line 123 which is `set: vi.fn(async () => undefined),`. My preference would be to use the modern async style, rather than directly using the Promise API. It does look a bit inconsistent in how functions that return promises are being defined.

ccp-bofai (approved):
LGTM

PR #158, opened 2026-08-11
refactor(extension): scheme-parameter sender guard

fcsondheim:
Hardens `isExtensionSender` (the fail-closed access guard behind every `extension`-only message route) to work on Firefox as well as Chrome, and adds test coverage. No behavior change on Chrome — the
guard still fails closed everywhere it did before; the only new trust is "URL under this browser's real own-origin," so the boundary check and the `getURL`-absent fallback are the parts worth a close read.

## Why

Every privileged route is gated by `hasRequiredSender` → `isExtensionSender`, which trusts a sender only if its URL starts with `chrome-extension://${browser.runtime.id}/`. On Firefox both the scheme (`moz-extension://`) and the host (a per-install UUID, not `runtime.id`) differ, so no extension sender matches and the guard rejects the extension's own pages. This isolates that fix ahead of the additive Firefox changes.

## Changes

- **`senderGuard.ts`** — derive the extension's own origin from `browser.runtime.getURL('/')` (the canonical scheme+host on both browsers: `chrome-extension://<id>` / `moz-extension://<uuid>`) instead of rebuilding it from `runtime.id`. Match sender URLs against that origin with a **boundary check** (`=== base || startsWith(base + '/')`) so a lookalike host (`…-evil`) can't prefix-escape. Fail-closed preserved: no metadata → `false`; when `getURL` is unavailable (non-extension context) it falls back to a best-effort extension-scheme allow, matching the previous degraded-path semantics. `isDappSender` unchanged.
- **New `__tests__/senderGuard.test.ts`** (16 tests). Chrome match, Firefox `moz-extension` match, the full fail-closed matrix (no metadata, `https` dApp, other-extension id, prefix-escape lookalike, moz-under-Chrome), and the `getURL`-unavailable fallback.
- **`messageHandler.test.ts`** — added `getURL` to its browser mock so the existing route-policy suite exercises the real origin-matching path rather than the fallback.

### DIFF
--- a/apps/extension/src/lib/background/__tests__/senderGuard.test.ts
+++ b/apps/extension/src/lib/background/__tests__/senderGuard.test.ts
@@ -1,0 +1,174 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import type { Browser } from 'wxt/browser'
+import { isDappSender, isExtensionSender } from '@/lib/background/senderGuard'
+
+type MsgSender = Browser.runtime.MessageSender
+
+const CHROME_ID = 'abcdefghabcdefghabcdefghabcdefgh'
+const FIREFOX_UUID = '11111111-2222-3333-4444-555555555555'
+
+/**
+ * Stub `browser.runtime.getURL` to model a given extension origin. Passing
+ * `undefined` models a context where getURL is unavailable (the fallback path).
+ */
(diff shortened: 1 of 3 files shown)


ccp-bofai (approved):
LGTM

PR #160, opened 2026-08-12
feat: firefox keeper

fcsondheim:
Chrome keeps the vault's ephemeral key in an offscreen document because its service worker can be suspended. Firefox has no offscreen API, so on Firefox the keeper runs in-process in the background instead. This completes the `KeeperHost` seam with a Firefox implementation selected at build time; the change to review is the transport bridge in `firefoxKeeperHost.ts`, which must reproduce Chrome's message semantics exactly.

### What changed
- **`FirefoxKeeperHost`** — `send()` dispatches directly to the same `handleKeeperMessage` router Chrome reaches over `runtime.sendMessage`, tagging `target: 'KEEPER'` and supplying an own-origin sender so the existing `isExtensionSender` guard passes unchanged. The `sendResponse` callback is bridged into a Promise, matching Chrome's contract (async/sync/no-reply, first response wins). `ensureReady()` is a no-op; `retries` is inert in-process.
- **Host selection** — `keeperHost` resolves via `selectKeeperHost(import.meta.env.FIREFOX)`: Firefox → in-process host, Chrome → offscreen host. Interface unchanged.
- **Manifest** — drop the now-unused `offscreen` permission from the Firefox build; Chrome keeps it.

### Testing
New `firefoxKeeperHost.test.ts` drives the real router (in-process routing, unknown-type error, no-context rejection, no-op `ensureReady`, both selection branches). Background + keeper suites green. Both targets build: Firefox manifest → `["identity","storage"]`, Chrome unchanged.

### Notes
- Chrome is untouched at runtime — same offscreen host and manifest.
- On Firefox the key lives in background-page memory; if the event page suspends, the user re-unlocks — the same UX as Chrome's service-worker suspension.

### DIFF
--- a/apps/extension/src/lib/background/handlers/sponsoredTransactionHandler.ts
+++ b/apps/extension/src/lib/background/handlers/sponsoredTransactionHandler.ts
@@ -151,7 +151,7 @@ async function handleSponsoredTransaction(
         assemblyType,
         metadata,
       },
-      sponsoredApiContext(apiBaseUrl, tenant, jwt.id_token, '/v2'),
+      sponsoredApiContext(apiBaseUrl, tenant, jwt.id_token, ''),
     )
 
     const actionType =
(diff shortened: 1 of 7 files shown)


ccp-bofai (approved):
LGTM

PR #162, opened 2026-08-17
feat(extension): Firefox async messaging fixes

fcsondheim:
Firefox has no offscreen API and rejects `onMessage` listeners that claim an async response but never send one — both required changes to the keeper and background messaging layers.

## Changes
- **Out-of-band message routes** — routes that reply via a separate message (`ext_login`, `connect`, sponsored, `web_unlock`) now close the channel instead of returning `true`, fixing Firefox's `Promised response … went out of scope` rejection (surfaced as an `AuthError`). `ext_login` reports unexpected failures out-of-band so the popup can't hang.
- **Login listener** — the id-correlated `extensionLogin` listener is now authoritative over a spurious send rejection.
- **Sponsored tx digest** — wallet-core 0.0.7 dropped `effects` from the execute result, so the content-script guard silently dropped `sign_success` and the digest never reached the dapp. Forward `{ digest, executionStatus }` end to end.

### DIFF
--- a/apps/extension/entrypoints/__tests__/content.test.ts
+++ b/apps/extension/entrypoints/__tests__/content.test.ts
@@ -468,6 +468,30 @@ describe('content bridge message validation', () => {
     )
   })
 
+  it('forwards a sponsored sign_success carrying digest and executionStatus but no effects', () => {
+    const postMessage = vi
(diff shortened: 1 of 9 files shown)


ccp-bofai (approved):
LGTM

PR #165, opened 2026-08-19
feat: add simulate transaction to approval popup

fcsondheim:
Addresses [EF-18338](https://fenriscreations.atlassian.net/browse/EF-18338)

When a dApp requests `signTransaction` or `signAndExecuteTransaction` (and for sponsored transactions), the approval popup previously showed only the decoded transaction payload — users had to reason about the raw PTB to know what would happen. This PR simulates the transaction against the fullnode **before approval** and surfaces the projected outcome, so users can see the concrete effect on their account instead of blind-approving.

### What's shown

The popup now leads with a **Projected outcome** view:

- **Status** — expected to succeed / expected to fail (with the on-chain abort reason)
- **Digest** — the projected transaction digest
- **Gas fee** — net (computation + storage − rebate), in SUI
- **Balance changes** — net per-coin deltas for the account, resolved to human-readable amounts/symbols via coin metadata
- **Changed objects** — each changed object's id, type, operation (created / mutated / deleted / published), and ownership transition (e.g. `you → 0x9f…aa`, `→ shared`)
- **Events** — emitted events with their decoded Move struct JSON (the contract's own description of what it did)

If the simulation predicts an **on-chain failure**, the user must tick the acknowledgement checkbox before approving (reuses the existing danger-acknowledgement gate). If simulation can't run (transport error), the popup shows a soft "could not simulate — approve with caution" notice with the reason, and does **not** hard-block.

### Screenshots

Screenshot 2026-08-19 at 15 58 33
Screenshot 2026-08-19 at 15 58 42

### How it works

- Reuses the gRPC client's `simulateTransaction` (already used for gas estimation) with `include: { effects, balanceChanges, objectTypes, events }`.
- The sender is resolved the same way the signing path does (`getSenderAddress`), not from the dApp payload — the wallet-standard `account.address` doesn't survive the storage round-trip into the popup.
- **dApp transactions** are built (`buildTransactionBytes`) then simulated, so the simulated bytes match what gets signed.
- **Sponsored transactions** are simulated from the already-built backend bytes as-is (rebuilding would drop the sponsor's gas payment).
- Simulation re-runs when the in-popup network selector changes, with a run-id guard to discard stale results.

### Scope

- Covers `sign_transaction`, `sign_and_execute_transaction`, and sponsored transactions.
- Personal-message signing is unaffected (nothing to simulate).

### DIFF
--- a/apps/extension/src/features/wallet/components/__tests__/SignTransactionFlow.test.tsx
+++ b/apps/extension/src/features/wallet/components/__tests__/SignTransactionFlow.test.tsx
@@ -1,12 +1,26 @@
 import { fireEvent, render, screen } from '@testing-library/react'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
-const { mockUseTransactionSigning } = vi.hoisted(() => ({
-  mockUseTransactionSigning: vi.fn(),
-}))
+const { mockUseTransactionSigning, mockUseTransactionSimulation } = vi.hoisted(
+  () => ({
+    mockUseTransactionSigning: vi.fn(),
+    mockUseTransactionSimulation: vi.fn(),
+  }),
+)
 
 vi.mock('@/features/wallet/hooks', () => ({
   useTransactionSigning: mockUseTransactionSigning,
+  useTransactionSimulation: mockUseTransactionSimulation,
+}))
+
+vi.mock('@evevault/shared/wallet', () => ({
+  useWalletSigningContext: () => ({ suiClient: {}, chain: 'sui:testnet' }),
+}))
+
+vi.mock('@/features/wallet/components/TransactionSimulationPanel', () => ({
+  TransactionSimulationPanel: ({ state }: { state: unknown }) => (
+    <div data-testid="simulation-panel">{JSON.stringify(state)}</div>
+  ),
 }))
 
 vi.mock('@/features/wallet/components/SignRequestView', () => ({
@@ -79,6 +93,7 @@ function stubSigning(overrides: Record<string, unknown> = {}) {
 
 beforeEach(() => {
   vi.clearAllMocks()
+  mockUseTransactionSimulation.mockReturnValue(null)
 })
 
 describe('SignTransactionFlow', () => {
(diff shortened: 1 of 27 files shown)


apps/extension/src/features/wallet/components/__tests__/SignTransactionFlow.test.tsx
```diff
@@ -135,6 +175,59 @@ describe('SignTransactionFlow', () => {
...
+      simulation: {
+        status: 'failure',
+        error: 'MoveAbort',
+        digest: 'd',
+        gas: { computation: '0', storage: '0', rebate: '0', net: '0.001' },
+        balanceChanges: [],
+        changedObjects: [],
+        events: [],
+      },
+    })
```
ccp-bofai, line 200: Is this realistic? as in does it reflect the reality of when you ask for a simulation that you do not have enough gas to cover that the API call returns an error not an result. I think it might actually raise an Error under the above circumstance further down.
Looking further down I think you might get a return with `status: 'unavailable', reason: ${err.message}` sort of thing. This is for a different error of course but I am not seeing any coverage of this type of error in the tests.

fcsondheim: This is a good flag, there is a `SimulationError` returned when there isn't enough gas. `status: 'unavailable'` was actually flawed as well as the true failure path wasn't being properly handled.

packages/shared/src/utils/format.ts
ccp-bofai, line 1: Not sure if this suggestion is a good idea or not :) but here is what I am noticing from this code. If you have the number 123.456 (en-US locale formatting). Then someone that is German may see the period as a thousands separator and read this as 123456.

I think `Intl.NumberFormat.format(...)` might be a way to deal with this. On [MDN it says that the `number` param can be a `BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#number). The table at the bottom of that page seems to report support for browser versions mainly from 2022 and some from 2023.

packages/shared/src/wallet/utils/simulateTransaction.ts
ccp-bofai, line 1: Should this move to Wallet-Core ?

ccp-bofai (approved):
Looks good to me... some comments to consider for moving forward

packages/shared/src/utils/format.ts
fcsondheim: Localisation of the display amount is handled by the function `formatDisplayAmount` - however you're right in that there were places in the UI where the numerical amount wasn't being parsed by the formatting function prior to being rendered on screen. I've wrapped those amounts in the display function

packages/shared/src/wallet/utils/simulateTransaction.ts
fcsondheim: Good call for a future PR!

PR #167, opened 2026-08-25
feat: enforce aliasing

fcsondheim:
Requires zkLogin accounts to register a recovery alias before signing **on-chain transactions**. Enforcement is always on. Suspending this requires an auditable, expiring ops override break-glass.

## Enforcement

- **Just-in-time gate (`useRequireAlias`)** — callers `await ensureAlias()` right before signing; it opens the recovery-setup modal only when an on-chain lookup finds no non-self alias. Wired into `SignTransactionFlow`, `SignSponsoredTransaction`, and `SendTokenScreen`.
- **Signing backstop (`assertAliasEnforced`)** — throws at the signing layer when no enforceable alias exists, so the policy holds even if a UI path misses the gate. Only on-chain transactions (`TransactionData`) are gated. Localnet is skipped.
- **On-chain state is the source of truth** — enforcement status is read live on every check (no client-side cache), so adding or removing an alias never requires cache invalidation and can't leave a stale "satisfied" result.
- **No kill-switch — auditable break-glass instead.** Enforcement is unconditional. The only escape is an ops-issued, **signed, expiring** override that is **fail-closed** (any malformed/expired/absent override → enforce) and logged on every use. The override validator, both call sites, and the audit logging are implemented; sourcing the override from a signed `alias_enforcement_override` JWT claim is stubbed as a fast-follow.

## Provisioning & setup UX

- **One-time client-only key** — generates a personal access key (24-word BIP-39 mnemonic + Ed25519 private key), shown once and never persisted; registered on-chain as the account's alias.

Screenshot 2026-08-27 at 11 52 03
Screenshot 2026-09-01 at 12 04 32

## Transaction history

- **Newest-first ordering** — the query paginated forward against an oldest-first connection. Switched to backward pagination (`last`/`before`).
- **Move-call counterparty labels** — coin-less transactions that only pay gas showed a bare `System`. They now read as `module::function` (e.g. `address_alias::add`); a real sender/recipient address still takes precedence.

## Follow-ups

- **Break-glass source:** wire `getEnforcementOverrideClaim()` to the signed `alias_enforcement_override` claim on the zkLogin JWT (backend change) so ops can engage the override without a client redeploy; validation, fail-closed handling, and audit logging are already in place.

### DIFF
--- a/packages/shared/src/wallet/__tests__/aliasEnforcement.test.ts
+++ b/packages/shared/src/wallet/__tests__/aliasEnforcement.test.ts
@@ -1,5 +1,5 @@
 import { SUI_DEVNET_CHAIN, SUI_LOCALNET_CHAIN } from '@mysten/wallet-standard'
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 const mockListOwnedObjects = vi.fn()
 
@@ -25,8 +25,8 @@ vi.mock('@mysten/sui/transactions', () => ({
 import { Transaction } from '@mysten/sui/transactions'
 import {
   assertAliasEnforced,
-  invalidateAliasEnforcement,
   isAliasEnforcementError,
+  resolveEnforcementOverride,
 } from '#/wallet/aliasEnforcement'
 
 const OWNER = `0x${'a'.repeat(64)}`
@@ -45,14 +45,8 @@ function stubDecodedModule(module: string) {
 }
 
 beforeEach(() => {
-  invalidateAliasEnforcement()
   mockListOwnedObjects.mockReset()
   vi.mocked(Transaction.from).mockReset()
-  vi.unstubAllEnvs()
-})
-
-afterEach(() => {
-  vi.unstubAllEnvs()
 })
 
 describe('assertAliasEnforced', () => {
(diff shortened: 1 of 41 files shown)


packages/shared/src/wallet/__tests__/aliasEnforcement.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    expect(resolveEnforcementOverride(null, NOW)).toBeNull()
+    expect(resolveEnforcementOverride('nope', NOW)).toBeNull()
+    expect(resolveEnforcementOverride({ until: NOW + 1 }, NOW)).toBeNull()
+    expect(
+      resolveEnforcementOverride({ reason: '', until: NOW + 1 }, NOW),
+    ).toBeNull()
+    expect(resolveEnforcementOverride({ reason: 'x' }, NOW)).toBeNull()
+    expect(
+      resolveEnforcementOverride({ reason: 'x', until: 'soon' }, NOW),
+    ).toBeNull()
```
ccp-bofai: The use of "rejects" in the test description is a bit confusing because of the Promises API and the `reject()` function. This is confusing when considering that the expectations are all `.toBeNull()`

ccp-bofai (approved):
LGTM but with a comment about wording of test descriptions

PR #171, opened 2026-09-02
fix: bind nonce to rotating ephkey

### LINKED ISSUE #126: High occurrence of `Failed to fetch ZK proof`
I am seeing a lot of `Failed to fetch ZK proof` that require me to re-propose transactions to the EveVault (v0.0.10).

This is the UX when this occurs:
1. DApp proposes a transaction to Eve Vault
2. Popup occurs from Eve Vault for signing a transaction
3. Click "approve" > Popup loads for 2-5 seconds, then displays `Failed to fetch ZK proof`
4. The popup doesn't close - it stays open.
5. Click "approve" again > Popup seems like it's loading for 2-5 seconds and then it closes.
6. Player has to re-trigger DApp-proposed transaction.
7. Click "approve" on new Popup > Transaction goes through.

This is happening ~40-50% of the time on DApp-proposed transactions at the moment. These DApp-proposed transactions utilize a Sui gas pool that i operate to sponsor transactions users create in my DApp.

fcsondheim:
fixes #126

zkLogin proof requests fail with Enoki `status 400: incorrect_nonce`. The `nonce`/`maxEpoch`/`jwtRandomness` triple is stored **per-chain** in `networkData`, but `ephemeralPublicKey` is a **global** field sourced from the keeper. Those can drift out of sync, leaving a stored nonce bound to a key the request no longer carries. `ensureProofNonce` only regenerated on *missing* or *epoch-expired* nonces, so a stale-but-unexpired nonce sailed straight through to a hard 400.

The keeper's unlock is deterministic decryption. Drift comes from the key and the nonce being written **non-atomically**: `rotateEphemeralKeyForChain` committed the new key with a blanked nonce, then `await`ed a second step (with a network call) to fill the nonce in. An interruption in that window persisted a new key with no/stale matching nonce.

### Changes on the branch

**1. Detection backstop — `proofHelpers.ts`**
`ensureProofNonce` now recomputes `generateNonce(currentKey, storedEpoch, storedRandomness)` and compares it to the stored nonce (`isNonceBoundToCurrentKey`). On mismatch it takes the existing rotate-and-regenerate path instead of sending a doomed request — the same check the backend does, moved client-side so drift self-heals. Wrapped in try/catch so malformed stored epoch/randomness also rotate rather than throw.

**2. Test fidelity — `deviceStore.proofActions.test.ts`**
The drift test now models rotation: the stub swaps `ephemeralPublicKey` to a new key and derives the nonce from it, assertions confirm the rotated key/nonce/randomness actually reach `resolveVendedIdTokenForZkProof` and `fetchZkProof` (including `ephemeralPublicKey`) — closing the gap where the test could pass with a mismatched key.

**3. Root-cause fix — atomic rotation in `initChainHelpers.ts`**
`rotateEphemeralKeyForChain` is now compute-then-commit: fetch the epoch *before* the irreversible keeper rotate, derive the nonce from the freshly-rotated key synchronously, and commit key + matching nonce in a **single `set()`**. The store can no longer persist a nonce bound to a superseded key. Extracted a pure `buildNonceEntry` helper reused by the normal init path.

### DIFF
--- a/apps/extension/src/lib/adapters/SuiWallet/__tests__/index.test.ts
+++ b/apps/extension/src/lib/adapters/SuiWallet/__tests__/index.test.ts
@@ -238,7 +238,7 @@ describe('EveVaultWallet', () => {
         },
       }),
     )
-    // The genuine same-window response still settles the connect.
+    // Same-window response still settles the connect.
     dispatchVaultMessage({
       type: 'auth_success',
       chain: SUI_LOCALNET_CHAIN,
--- a/packages/shared/src/stores/__tests__/deviceStore.proofActions.test.ts
+++ b/packages/shared/src/stores/__tests__/deviceStore.proofActions.test.ts
@@ -1,4 +1,5 @@
 import { Ed25519PublicKey } from '@mysten/sui/keypairs/ed25519'
+import { generateNonce } from '@mysten/sui/zklogin'
 import { SUI_DEVNET_CHAIN } from '@mysten/wallet-standard'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { createProofActions } from '#/stores/deviceStore/actions/proofActions'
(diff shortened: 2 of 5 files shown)


ccp-bofai (approved):
LGTM

PR #172, opened 2026-09-02
refactor: use simulate tx from `wallet-core`

fcsondheim:
`@evefrontier/wallet-core` [PR #55](https://github.com/evefrontier/wallet-core/pull/55) upstreamed the transaction-simulation logic that originally lived in this repo. This PR deletes the local copies and consumes the library instead, keeping evevault as the single source only for the pieces that are genuinely app-specific (the React hook, the GraphQL metadata source).

### Changes

**Simulation core → wallet-core**
- Deleted `packages/shared/src/wallet/utils/simulateTransaction.ts` (the full local port: `simulateTransactionOutcome`, `classifyBuildFailure`, and the `Simulated*`/`TransactionSimulation` types).
- Deleted its test `simulateTransaction.test.ts` — this logic is now covered by wallet-core's own test suite.
- `packages/shared/src/wallet/index.ts` re-exports the simulation symbols from `@evefrontier/wallet-core/transaction`, and now also exports `fetchCoinMetadata` so it can be injected.

**Formatters → wallet-core**
- Removed the local `formatByDecimals` / `formatMistToSui` from `utils/format.ts` (kept the un-ported `toSmallestUnit`, `formatDisplayAmount`, `formatShortDate`).
- Updated call sites to import both from `@evefrontier/wallet-core/utils`: `useBalance.helpers.ts`, `useSendToken.gas.ts`, `parseTransaction.ts`, and the `format.node.test.ts` / `useBalance.test.tsx` / `useSendToken.gas.test.ts` tests.

**Call-site wiring**
- `useTransactionSimulation.ts`: wallet-core's `simulateTransactionOutcome` takes an injected `resolveCoinMetadata` resolver instead of a raw `graphqlClient`, so the library stays free of evevault's GraphQL layer. The hook now passes a resolver that calls evevault's existing cached `fetchCoinMetadata`, normalizing its `name: string | null` to the `string | undefined` the library's `CoinMetadata` type expects. Same client, same cache, same fallback behavior.

**Dependencies**
- Bumped `@evefrontier/wallet-core` `0.0.7` → `0.0.8`.
- Added `fast-uri` (pulled in by the updated dependency tree).


ccp-bofai (approved):
LGTM

PR #174, opened 2026-09-04
chore: use wallet-core coin cache

fcsondheim:
Upgrades `@evefrontier/wallet-core` to `0.0.11` and moves coin-metadata caching out of evevault into wallet-core's `CachedCoinMetadataResolver`. wallet-core now owns the cache, its 30-minute TTL, and the known-coin (SUI/EVE) shortcuts; evevault supplies only the GraphQL lookup.

### Changes
- **`packages/shared/.../utils/coinMetadata.ts`** — removed the local `Map` cache, TTL constant, and SUI-only known-coin handling. Added `createGraphQLCoinMetadataResolver(graphqlClient)`, which wraps evevault's GraphQL query as the resolver's lookup. `fetchCoinMetadata`/`invalidateCoinMetadataCache` are retained (used by transaction-history parsing) and now delegate to one resolver per GraphQL client.
- **`apps/extension/.../useTransactionSimulation.ts`** — builds the resolver once per network via `useMemo(chain)` and passes `resolver.resolve` straight into `simulateTransactionOutcome`, instead of constructing an inline metadata mapper on every run. `description`/`iconUrl` now flow through to simulated balance changes.
- **Types/exports** — deleted the now-redundant `CoinMetadataResult` interface (replaced by wallet-core's `ResolvedCoinMetadata`); exported `createGraphQLCoinMetadataResolver`, `CachedCoinMetadataResolver`, and `ResolvedCoinMetadata`.

### Behavior change
The previous local cache was keyed by coin type across all networks. Caches are now isolated per GraphQL client (per network), so the same coin type on different networks no longer shares an entry — more correct, since metadata can differ across networks.

### DIFF
--- a/apps/extension/src/features/wallet/hooks/__tests__/useTransactionSimulation.test.ts
+++ b/apps/extension/src/features/wallet/hooks/__tests__/useTransactionSimulation.test.ts
@@ -7,14 +7,21 @@ const {
   mockSimulateTransactionOutcome,
   mockClassifyBuildFailure,
   mockCreateSuiGraphQLClient,
-  mockFetchCoinMetadata,
-} = vi.hoisted(() => ({
-  mockBuildTransactionBytes: vi.fn(),
-  mockSimulateTransactionOutcome: vi.fn(),
-  mockClassifyBuildFailure: vi.fn(),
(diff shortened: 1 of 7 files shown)


ccp-bofai (approved):
LGTM

PR #175, opened 2026-09-04
feat: guard address alias calls from dapps

fcsondheim:
Alias transactions must only ever be driven from Eve Vault's own UI, never signed on from a third-party dApp. This PR adds that guard at two layers and closes a gap in the existing enforcement exemption.

### 1. `feat: alias guard UI` — user-facing gate
- Detects when a pending transaction (regular or sponsored) contains an `address_alias` MoveCall and disables **Approve**, showing a warning banner while still letting the user review and **Reject**.
- Defaults the approval view to the **Warnings** tab when a request is blocked.
- Detection lives in one place (`packages/shared/src/wallet/aliasCall.ts`, `transactionContainsAddressAliasCall`) and is re-exported for the extension.

Screenshot 2026-09-04 at 16 37 50

Also folds in two correctness fixes to that flow:
- The sponsored payload is now decoded via `b64ToBytesOrNull`, so a malformed base64 payload blocks Approve instead of throwing during render.
- `initialId` no longer points at a `warnings` tab that may not exist.

### 2. `feat: enforce address alias guard` — signing-layer backstop
- `signForChain` refuses any `TransactionData` carrying an alias call, on all chains, before the button is ever involved.
- `sign` gained an opt-in `{ allowAddressAliasCalls }`; only Eve Vault's own alias flows (`useAddressAliases`, `useAliasProvisioning`) set it. Every dApp path — normal signing, sponsored, and the SuiWallet adapter — is refused by default. Detection fails open here so a parser hiccup can't block unrelated signatures.
- Tightens the enforcement setup-exemption from `.some` to `.every`: only a *pure* alias-setup transaction is exempt, so a dApp can't bundle an alias call with other commands to dodge enforcement. Pure single- or multi-command alias setups (including first-time passkey `enable` + `add`) remain exempt.

## Behavior

| Origin                         | Contains alias call | Result                                       |
| ------------------------------ | ------------------- | -------------------------------------------- |
| dApp (any signing path)        | yes                 | Approve disabled + refused at `signForChain` |
| dApp                           | no                  | signs normally                               |
| Eve Vault alias setup/recovery | yes                 | signs (opted in)                             |

### DIFF
--- a/apps/extension/src/features/wallet/__tests__/aliasCallGuard.test.ts
+++ b/apps/extension/src/features/wallet/__tests__/aliasCallGuard.test.ts
@@ -1,0 +1,54 @@
+import { addAddressAliasTx } from '@evefrontier/wallet-core/address-alias'
+import { Transaction } from '@mysten/sui/transactions'
+import { describe, expect, it } from 'vitest'
+import { transactionContainsAddressAliasCall } from '../aliasCallGuard'
+
+const OWNER = `0x${'a'.repeat(64)}`
+const ALIASES_OBJECT = `0x${'b'.repeat(64)}`
+const ALIAS = `0x${'c'.repeat(64)}`
+const COIN = `0x${'d'.repeat(64)}`
+
+describe('transactionContainsAddressAliasCall', () => {
+  it('is true for a transaction whose only command is an alias call', async () => {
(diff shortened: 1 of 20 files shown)


ccp-bofai (approved):
LGTM but some of this should probably move to wallet-core.

PR #179, opened 2026-09-14
fix: web PIN keypair IndexedDB ejection

fcsondheim:
On the web app, returning after being away for a while and entering your PIN could dead-end with an unrecoverable lock screen.

## Root cause

The web wallet splits state across two browser stores with different durability:

- **IndexedDB** holds the ephemeral keypair and PIN verifier.
- **localStorage** holds the "this device is configured" markers.

Browsers can evict IndexedDB while leaving localStorage intact (Safari/iOS ITP after idle days, Chromium eviction under storage pressure). When that happens, localStorage still says "configured," so the app shows the **"Enter pin"** lock screen — but there's no keypair behind it, and every unlock attempt fails.

## Fix

The PIN is only a local unlock gate, So we treat eviction as an expired session:

1. **Detect at rehydration** — when the localStorage marker says "configured" but the IndexedDB keypair is genuinely gone, clear local state and route to login with a `sessionExpired` flag. Detection only triggers on a clean "key absent" result; a transient IndexedDB-open failure throws and leaves the user locked (never signed out).
2. **Explain the sign-out** — the login screen shows a short notice ("this browser cleared your local session… your wallet and address are unchanged") instead of a confusing error or an unprompted logout. The user signs in again (a silent SSO redirect when available) and sets a PIN through the normal onboarding flow.
3. **Reduce how often it happens** — request `navigator.storage.persist()` on web startup so the origin's storage is less likely to be evicted (best-effort; the recovery path still handles eviction when it occurs).

Also bundled: dependency updates.


ccp-bofai (approved):
LGTM

